PR #6, opened 2023-01-07
Erase `TestScheduler`'s value in the documentation

tgrapperon:


mbrandonw (approved):
Good catch, thanks!

I also noticed that TimerModel is a strict above. Mind changing that to a class too?

tgrapperon:
@mbrandonw Done!

PR #7, opened 2023-01-07
Create assert dependency

john-flanagan:
What do you think of adding an `assert` dependency to allow for testable assertions?

If you like the idea I can flesh out the documentation and add `assertionFailure`

### DIFF
--- a/Sources/Dependencies/DependencyValues/Assert.swift
+++ b/Sources/Dependencies/DependencyValues/Assert.swift
@@ -1,0 +1,35 @@
+import XCTestDynamicOverlay
+
+public struct AssertAction {
+  public let action: (@autoclosure () -> Bool, @autoclosure () -> String, StaticString, UInt) -> ()
+
+  public init(action: @escaping (() -> Bool, () -> String, StaticString, UInt) -> Void) {
+    self.action = action
+  }
+
+  @inline(__always)
+  public func callAsFunction(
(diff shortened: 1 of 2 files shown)


john-flanagan:
@stephencelis @mbrandonw What do you think of this addition?

stephencelis:
@john-flanagan Thanks for the PR, this is fun! I think we'll be a little conservative adding too many core dependencies to the library for now. Ideally we figure out an internal modularization strategy that makes it a bit easier to do so.

We definitely think it's worth sharing this on the GitHub discussions, though, and if you ship a library that contains this dependency and/or more, we'll definitely link to it in the README. You may also chat with @tgrapperon to see if he thinks it could belong in his [Dependencies Additions](https://github.com/tgrapperon/swift-dependencies-additions) extension library.

tgrapperon:
Hey @john-flanagan! I'm indeed totally open to include this extension in "Dependencies Additions". I'm setting up the Discussions page of the repo, so feel free to open a discussion there or to PR if you want. There are safety precaution to take though, as this dependency could easily be exploited to bypass internal checks. But there are ways to harden them, so we can discuss!

john-flanagan:
@tgrapperon I opened a discussion on your repo! https://github.com/tgrapperon/swift-dependencies-additions/discussions/17

Looks like you already have a lot of great additions.

PR #10, opened 2023-01-10
Incrementing date generator

runloop:
This is a `DateGenerator` based off the `UUIDGenerator.incrementing`. It provides a predictable sequence of dates allowing you to test time changes.

### DIFF
--- a/Sources/Dependencies/DependencyValues/Date.swift
+++ b/Sources/Dependencies/DependencyValues/Date.swift
@@ -34,7 +34,7 @@ extension DependencyValues {
     get { self[DateGeneratorKey.self] }
     set { self[DateGeneratorKey.self] = newValue }
   }
-
+  
   private enum DateGeneratorKey: DependencyKey {
     static let liveValue = DateGenerator { Date() }
     static let testValue = DateGenerator {
@@ -49,7 +49,7 @@ extension DependencyValues {
 /// See ``DependencyValues/date`` for more information.
 public struct DateGenerator: Sendable {
   private var generate: @Sendable () -> Date
-
+  
   /// A generator that returns a constant date.
   ///
   /// - Parameter now: A date to return.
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hi @runloop, thanks for taking the time to contribute! Can you provide some more context of how one would use this in practice?

We think such a thing is probably best handled using clocks or Combine schedulers. For example, `RunLoop`s conform to the `Scheduler` protocol, and its `now` property is a `Date`. So you can do things like this in your feature:

```swift
@Dependency(\.mainRunLoop) var mainRunLoop
for await date in mainRunLoop.timer(interval: .seconds(1)) {
  self.label = "It is \(date)"
}
```

Then in tests you can do:

```swift
let mainRunLoop = RunLoop.test
let model = withDependencies { 
  $0.mainRunLoop = mainRunLoop.eraseToAnyScheduler()
} operation: { 
  FeatureModel()
}

Task {
  await model.onAppear()
}

mainRunLoop.advance(by: .seconds(1))
XCTAssertEqual(model.label, "It is 2009-02-13 23:31:31 +0000")
mainRunLoop.advance(by: .seconds(1))
XCTAssertEqual(model.label, "It is 2009-02-13 23:31:32 +0000")
mainRunLoop.advance(by: .seconds(1))
XCTAssertEqual(model.label, "It is 2009-02-13 23:31:33 +0000")
mainRunLoop.advance(by: .seconds(1))
XCTAssertEqual(model.label, "It is 2009-02-13 23:31:34 +0000")
```

This makes it so that the flow of time is inextricably linked with the scheduling of work.

You could even define a computed property on `DependencyValues` that computes a "now" from `mainRunLoop.now`. The only reason we don't do that in the library is because all the Combine stuff is closed source and so we don't want to tie ourselves to it too much.

In the future there is supposed to be a `UTCClock` with an instant of `Date`, and so once we have that and once we figure out some difficulties around the `Instant` type being erased in `Clock` we may try redefining `@Dependency(\.date)` to pull from `@Depenency(\.utcClock)`.

runloop:
No worries. However, your example is much better and deals with the issue I was already facing about linking time with the scheduling.

PR #11, opened 2023-01-10
Capture current dependencies when escaping

tgrapperon:
This should fix https://github.com/pointfreeco/swift-composable-architecture/issues/1811.
I don't if this is the behavior you intend for `DependencyValues.Continuation` overall, so feel free to close if we should fix it in TCA instead.

It would perform mostly equivalently to the current implementation in vanilla applications, where nesting is less intense than in TCA, especially if the continuation is yielded as soon as possible in the escaped context.

### DIFF
--- a/Sources/Dependencies/WithDependencies.swift
+++ b/Sources/Dependencies/WithDependencies.swift
@@ -353,6 +353,7 @@ extension DependencyValues {
     /// See the docs of ``withEscapedDependencies(_:)-5xvi3`` for more information.
     /// - Parameter operation: A closure which will have access to the propagated dependencies.
     public func yield<R>(_ operation: () throws -> R) rethrows -> R {
+      // TODO: Should `yield` be renamed to `restore`?
       try withDependencies {
         $0 = self.dependencies
       } operation: {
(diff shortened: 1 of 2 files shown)


Sources/Dependencies/WithDependencies.swift
```diff
@@ -338,15 +338,15 @@ public func withEscapedDependencies<R>(
...
+  try await operation(DependencyValues.Continuation(dependencies: ._current))
 }
 
 extension DependencyValues {
   /// A capture of dependencies to use in an escaping context.
   ///
   /// See the docs of ``withEscapedDependencies(_:)-5xvi3`` for more information.
   public struct Continuation: Sendable {
-    @Dependency(\.self) private var dependencies
+    let dependencies: DependencyValues
```
stephencelis, line 349: Could also localize the capture here:

```suggestion
    let dependencies = DependencyValues._current
```

Sources/Dependencies/WithDependencies.swift
```diff
@@ -338,15 +338,15 @@ public func withEscapedDependencies<R>(
 public func withEscapedDependencies<R>(
   _ operation: (DependencyValues.Continuation) async throws -> R
 ) async rethrows -> R {
-  try await operation(DependencyValues.Continuation())
+  try await operation(DependencyValues.Continuation(dependencies: ._current))
```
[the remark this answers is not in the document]
tgrapperon: I'm pushing more stuff.

Sources/Dependencies/WithDependencies.swift
tgrapperon: I'll rework the latest commit in this direction.

mbrandonw (approved):
awesome, thanks so much @tgrapperon!

PR #26, opened 2023-01-17
Remove unnecessary struct extension

oronbz:
This may be a leftover or an oversight, but correct me if I'm wrong, this docs part is unrelated, and the `DependencyKey` `extension` should be enough.

### DIFF
--- a/Sources/Dependencies/Documentation.docc/Articles/DesigningDependencies.md
+++ b/Sources/Dependencies/Documentation.docc/Articles/DesigningDependencies.md
@@ -87,12 +87,7 @@ extension AudioPlayerClient {
 
   static let mock = Self(/* ... */)
 
-  static let unimplemented = Self(
-    loop: unimplemented("AudioPlayerClient.loop"),
-    play: unimplemented("AudioPlayerClient.play"),
-    setVolume: unimplemented("AudioPlayerClient.setVolume"),
-    stop: unimplemented("AudioPlayerClient.stop")
-  )
+  static let unimplemented = Self(/* ... */)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @oronbz, the docs may just be a little confusing here, but technically that passage is intended.

It's trying to demonstrate the differences between the struct-style of dependencies and the protocol-style. In the struct-style you construct values to represent instances of the dependency, whereas in the protocol-style you create types that conform.

that when one uses structs to model dependencies, in order to construct "instances" of the dependency you construct values of the struct rather than create types that conform to a protocol.

I'll think of ways to make that clearer.

oronbz:
Yes, but it seems both of those sections:
```Swift
extension AudioPlayerClient {
  static let live = Self(/* ... */)
  static let mock = Self(/* ... */)
  static let unimplemented = Self(/* ... */)
}
```

```Swift
extension AudioPlayerClient: DependencyKey {
  static let liveValue = Self(/* ... */)
  static let previewValue = Self(/* ... */)
  static let testValue = Self(/* ... */)
}
```
Are showing the exact same point that you mentioned, while the second one will satisfy the default dependencies which is probably the best (better?) practice.

mbrandonw:
Let me take a quick pass at fleshing out what I had in mind for that passage. I do want to be a little more explicit.

mbrandonw:
Hi @oronbz, I just pushed an edit to your branch. Can you let me know if that clarifies things?

I also took your suggestion to add the `DependencyValues` computed property. Thanks!

oronbz:
@mbrandonw I was about to respond that it was even worse now IMO, but after the latest commit, it seems a bit clearer now, with more examples.

oronbz:
Also, thinking out loud here, but I wonder how to incorporate the `@Dependency` inside a `Dependency` example that I gave here: https://github.com/pointfreeco/swift-dependencies/pull/27/files to one of the main topics, maybe here using AVAudioEngine as a `dependency`?

It was the first thing that I've tried to look for without any luck, and pretty common use case in any DI library/solution.

mbrandonw:
We have plans to write up more documentation about that pattern. We left it out because we wanted to understand the ins and outs, but we now have good test coverage on it and we use that style out in the wild.

I will merge this PR as-is and we will get to that topic sometime soon.

PR #27, opened 2023-01-17
Align URLSession doc to benefit from `liveValue`.

oronbz:
This is a proposal to to show an example of how to incorporate a dependency within a dependency while still benefiting from the auto `liveValue` injection, without the need to manually inject it with the `.depencency(...)` modifier.


mbrandonw (approved):
Nice catch, this is a great change. Thanks!

PR #29, opened 2023-01-18
Fix default value fallback for Optional-valued dependencies

pyrtsa:
The `liveValue`/`previewValue`/`testValue` fallback mechanism is currently bypassed for dependencies which have an `associatedtype Value` of `Optional<_>` type. The logic confuses no value found in  `storage` for a valid value (`nil: Value`) found, and never ends up picking up the task-local default:

```swift
extension DependencyValues {
  var optionalDependency: String? {
    get { self[OptionalDependencyKey.self] }
    set { self[OptionalDependencyKey.self] = newValue }
  }
}

private enum OptionalDependencyKey: DependencyKey {
  static let liveValue: String? = "live"  // ⚠️ never returned!
  static var testValue: String? {         // ⚠️ never evaluated!
    unimplemented(#"@Dependency(\.optionalDependency)"#, placeholder: nil)
  }
}

// ⚠️ defaults `nil` no matter what!
@Dependency(\.optionalDependency) var optionalDependency: String?
```

This PR adds a a few unit test cases exposing the faulty behaviour and makes the fixes to correct it.

### DIFF
--- a/Tests/DependenciesTests/DependencyValuesTests.swift
+++ b/Tests/DependenciesTests/DependencyValuesTests.swift
@@ -71,6 +71,43 @@ final class DependencyValuesTests: XCTestCase {
     }
   }
 
+  func testOptionalDependency() {
+    for value in [nil, ""] {
+      withDependencies {
+        $0.optionalDependency = value
+      } operation: {
+        @Dependency(\.optionalDependency) var optionalDependency: String?
+        XCTAssertEqual(optionalDependency, value)
+      }
+    }
+  }
+
+  func testOptionalDependencyLive() {
+    withDependencies {
+      $0.context = .live
+    } operation: {
+      @Dependency(\.optionalDependency) var optionalDependency: String?
+      XCTAssertEqual(optionalDependency, "live")
+    }
+
+    withDependencies {
+      $0.context = .live
+      $0.optionalDependency = nil
+    } operation: {
+      @Dependency(\.optionalDependency) var optionalDependency: String?
+      XCTAssertNil(optionalDependency)
+    }
+  }
+
+  func testOptionalDependencyUndefined() {
+    @Dependency(\.optionalDependency) var optionalDependency: String?
+    XCTExpectFailure {
+      XCTAssertNil(optionalDependency)
+    } issueMatcher: {
+      $0.compactDescription.contains(#"Unimplemented: @Dependency(\.optionalDependency) …"#)
+    }
+  }
+
   func testDependencyDefaultIsReused() {
     withDependencies {
       $0 = .init()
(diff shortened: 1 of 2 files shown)


Tests/DependenciesTests/DependencyValuesTests.swift
```diff
@@ -71,6 +71,43 @@ final class DependencyValuesTests: XCTestCase {
...
+  }
+
+  func testOptionalDependencyUndefined() {
+    @Dependency(\.optionalDependency) var optionalDependency: String?
+    XCTExpectFailure {
+      XCTAssertNil(optionalDependency)
+    } issueMatcher: {
+      $0.compactDescription.contains(#"Unimplemented: @Dependency(\.optionalDependency) …"#)
+    }
+  }
```
stephencelis, line 109: Core libs XCTest unfortunately doesn't support `XCTExpectFailure`.

```suggestion
  #if !os(Linux)
  func testOptionalDependencyUndefined() {
    @Dependency(\.optionalDependency) var optionalDependency: String?
    XCTExpectFailure {
      XCTAssertNil(optionalDependency)
    } issueMatcher: {
      $0.compactDescription.contains(#"Unimplemented: @Dependency(\.optionalDependency) …"#)
    }
  }
  #endif
```

pyrtsa:
> Great catch, thanks! And thanks for the test coverage 😄

Oh, I see. Thanks for fixing it.

stephencelis:
@pyrtsa This one looks like potentially a real failure? Got time to look into it?

pyrtsa:
I'll check it out in a bit.

Tests/DependenciesTests/DependencyValuesTests.swift
```diff
@@ -71,6 +71,45 @@ final class DependencyValuesTests: XCTestCase {
...
+
+  #if !os(Linux)
+  func testOptionalDependencyUndefined() {
+    @Dependency(\.optionalDependency) var optionalDependency: String?
+    XCTExpectFailure {
+      XCTAssertNil(optionalDependency)
+    } issueMatcher: {
+      $0.compactDescription.contains(#"Unimplemented: @Dependency(\.optionalDependency) …"#)
+    }
+  }
```
pyrtsa, line 110: Hmm, looks like it was just due to `XCTestDynamicOverlay.XCTFail` becoming no-op in release builds.
```suggestion
  #if DEBUG && !os(Linux)
  func testOptionalDependencyUndefined() {
    @Dependency(\.optionalDependency) var optionalDependency: String?
    XCTExpectFailure {
      XCTAssertNil(optionalDependency)
    } issueMatcher: {
      $0.compactDescription.contains(#"Unimplemented: @Dependency(\.optionalDependency) …"#)
    }
  }
```

pyrtsa:
@stephencelis Could you give the tests another spin now?

stephencelis:
Great! Thanks again!

PR #30, opened 2023-01-19
Test values should default to preview before live

stephencelis:
If you define a live value and preview value, we should probably default to the preview value in tests to avoid accidentally hitting the live implementation. The test will fail when the dependency is accessed regardless.


mbrandonw (approved):
Good catch. I wonder if we had it like this at some point and then lost it in the port over to the new repo, because the docs describe it exactly like this:

https://pointfreeco.github.io/swift-dependencies/main/documentation/dependencies/livepreviewtest#Cascading-rules

PR #31, opened 2023-01-20
Use early return in DependencyValues

timothycosta:
This is just a minor cleanup of of DependencyValues.swift to use early return and factor out the warning in order to make it easier to understand the logic of how values get returned.

I hope this sort of PR would be ok. I was browsing the source just because I found this project interesting. I'm not actually using it atm.😅


stephencelis:
@timothycosta Thanks for sharing!

I think we personally are a bit more comfortable with the existing `guard`s that we're using and how the compiler enforces returns on those code paths, and I think merging a change here would make it a lil harder on us as maintainers to follow. Because of this, I'm going to close the PR for now, but I do think it's always worth looking at and considering alternatives, so thanks again!

PR #34, opened 2023-01-22
Correct LivePreviewTest link; remove random characters

acosmicflamingo:
The Testing.md in its current state has a missing link, and I don't know what to do about the random characters appearing after "withDependencies(_:operation:)", so I just deleted them.

### DIFF
--- a/Sources/Dependencies/Documentation.docc/Articles/Testing.md
+++ b/Sources/Dependencies/Documentation.docc/Articles/Testing.md
@@ -5,7 +5,7 @@ tricks for writing better tests with the library.
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


tgrapperon:
These characters are automatically filed by Xcode to disambiguate overloads. They can be out of sync sometimes, but I would put them back again, as it seems it's not the case here.
Furthermore, it seems to be working correctly here: https://pointfreeco.github.io/swift-dependencies/main/documentation/dependencies/testing. Is this not working from Xcode?

acosmicflamingo:
Ah I see...I feel silly now. I was accessing the documentation directly through the README.md in the repo; that link is where I should actually be going :|

tgrapperon:
Don't feel silly it happens to everyone! You now have a better way to read the doc, so it's globally a win!

PR #51, opened 2023-02-19
Add a reference to Weaver to the "Alternatives" section.

tgrapperon:
GitHub just suggested me this 700 stars dependency-injection repo that I wasn't aware of. It looks rather complex, with some bits of codegen, but it also uses modern Swift features like property wrappers. I seems to fit in this section.


mbrandonw (approved):
Yep, seems good to me. Thanks!

PR #53, opened 2023-02-23
Add a `isEmpty` property to `DependencyValues`

tgrapperon:
This PR adds an `isEmpty` property to `DependencyValues` that reflects the current storage's state. 
This property can be used to detect if the current context has dependencies with non-default values, as it was potentially discussed on [Slack](https://pointfreecommunity.slack.com/archives/C04L2D0MNJH/p1676982334184199).
This property, which reflects some internal implementation state, is hidden behind `@_spi`, so it furthermore doesn't appear in `@Dependency(\.…)` `KeyPath` autocompletion.
As an alternative, it can directly be installed as a static property that would only reflect `Self._current`'s internal storage state. Let me know if you prefer it this way instead.

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -85,6 +85,7 @@ public struct DependencyValues: Sendable {
   @TaskLocal static var currentDependency = CurrentDependency()
 
   fileprivate var cachedValues = CachedValues()
+  @_spi(Internals) public var isEmpty: Bool { self.storage.isEmpty }
   private var storage: [ObjectIdentifier: AnySendable] = [:]
 
   /// Creates a dependency values instance.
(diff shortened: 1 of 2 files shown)


tgrapperon:
Parking as a draft for now.

stephencelis:
I'm going to close this for now, but we could maybe revisit in the future with more discussion an context (sadly, the Slack history no longer goes as far back as the link, so if you remember what was said please do include it here or in a future discussion/PR!)

tgrapperon:
IRRC, the objective was to allow to not inject dependencies if they all had their default values when a Store was created within a `WithDependencies` block. `isEmpty` was a way to detect that they were no overrides. I don't think it would have saved much in the cased it would have applied. I forgot about this PR, but I should have closed it earlier.

PR #54, opened 2023-02-23
Modernize a few tests

tgrapperon:
Xcode 14.3 shows only a few new warnings that seem easy to fix.
On occurrence will need `import Foundation` in Swift 6 to conform `String` to `CVarArg`.
The other changes are upgrades from `wait(for: expectations)` to `await fulfillment(of: expectations)` on Apple platforms.


tgrapperon:
Ha, `fulfillment` APIs are annotated iOS 13 & co, but they're not available on Xcode 14.2 so I suppose that they back-deploy,
I'm closing the PR for now, as this version is still in beta.

PR #56, opened 2023-02-27
Document the issue about linking `Dependencies` in the tests target

tgrapperon:
There were recently a few reports of users with this sneaky issue. Feel free to fix the wording/typos.


stephencelis (approved):
Thanks! Will be handy to have an explicit link for folks in the future.

PR #58, opened 2023-03-05
Simplify the `OpenURLEffect.liveValue` implementation.

zmian:
# Changelog
- Simplify the `OpenURLEffect.liveValue` implementation using `withCheckedContinuation`.
- Added `OpenURLTests`

Hello, While looking at the implementation of `OpenURLEffect` and noticed that it's using the `AsyncStream` and grabbing the first element and discarding rest. In this instance, `withCheckedContinuation` should be sufficient to covert the closure into async API.

Thanks for taking a look.

### DIFF
--- a/Sources/Dependencies/DependencyValues/OpenURL.swift
+++ b/Sources/Dependencies/DependencyValues/OpenURL.swift
@@ -12,26 +12,19 @@
 
   @available(iOS 14, macOS 11, tvOS 14, watchOS 7, *)
   private enum OpenURLKey: DependencyKey {
-    static let liveValue = OpenURLEffect { url in
-      let stream = AsyncStream<Bool> { continuation in
-        let task = Task { @MainActor in
-          #if os(watchOS)
-            EnvironmentValues().openURL(url)
-            continuation.yield(true)
-            continuation.finish()
-          #else
-            EnvironmentValues().openURL(url) { canOpen in
-              continuation.yield(canOpen)
-              continuation.finish()
-            }
-          #endif
+    static let liveValue = OpenURLEffect { @MainActor url in
+      #if os(watchOS)
+        EnvironmentValues().openURL(url)
+        return true
+      #else
+        return await withCheckedContinuation { continuation in
+          EnvironmentValues().openURL(url) { canOpen in
+            continuation.resume(returning: canOpen)
+          }
         }
-        continuation.onTermination = { @Sendable _ in
-          task.cancel()
-        }
-      }
-      return await stream.first(where: { _ in true }) ?? false
+      #endif
     }
+
     static let testValue = OpenURLEffect { _ in
       XCTFail(#"Unimplemented: @Dependency(\.openURL)"#)
       return false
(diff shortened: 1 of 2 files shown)


stephencelis:
Hey @zmian! Thanks for taking the time to PR.

We typically prefer wrapping callback work with `AsyncStream` over `withCheckedContinuation` because `AsyncStream` participates in cooperative cancellation, whereas `withCheckedContinuation` does not. I'm not actually sure it matters in this particular case, though I do worry about causing cancelled async work to hang longer than it should with changes like this.

Do you have any insight as to when we should prefer `withCheckedContinuation` when we want cancellation to work quickly from outside the task?

zmian:
Hi @stephencelis, thanks for taking a look and brining up the cooperative cancellation. 

As you've noted `withCheckedContinuation` does not participate in cooperative cancellation. To enable cooperative cancellation in `withCheckedContinuation` and family it must be wrapped in `Task.withCancellationHandler` to manually propagate the outside task cancellation to the inner operation. [Please see here for some more info around this in the original proposal.](https://github.com/apple/swift-evolution/blob/main/proposals/0300-continuation.md#additional-examples)

In this particular instance the cancellation is not really needed. However, I do understand the desire to keep things in consistent manner within the library. 

We have few options:
1. I can revert the `withCheckedContinuation` change. And, only keep the tests if you are happy with them.
2. Create a lightweight internal function that wraps `AsyncStream` with similar  API to `withCheckedContinuation`. This allows the complexity to live in this simple function and makes reuse simpler if it's needed else where. 
3. Happy to close this pr.

zmian:
Hi @stephencelis, thanks for taking a look and brining up the cooperative cancellation. 

As you've noted `withCheckedContinuation` does not participate in cooperative cancellation. 

To enable cooperative cancellation _like behavior_ in `withCheckedContinuation` and family it must be wrapped in `withTaskCancellationHandler` to manually propagate the outside task cancellation to the inner operation. This will get you close enough with slight differences from the way cooperative cancellation works in the structured concurrency. Which makes sense since `withCheckedContinuation` is meant for bringing over the completion handlers and delegate methods where cancellation concept is context dependent. Some of this behavior is outlined [here](https://github.com/apple/swift-evolution/blob/main/proposals/0300-continuation.md#additional-examples) in the original proposal and [here](https://github.com/apple/swift/blob/5665705f16a8d052fecc46da7f0c0a0a34c6c101/stdlib/public/Concurrency/TaskCancellation.swift#L18-L38).

In this particular instance the cancellation is not really needed. However, I do understand the desire to keep things in consistent manner within the library. 

We have few options:
1. I can revert the `withCheckedContinuation` change. And, only keep the tests if you are happy with them.
2. Create a lightweight internal function that wraps `AsyncStream` with similar  API to `withCheckedContinuation`. This allows the complexity to live in this simple function and makes reuse simpler if it's needed else where. 
3. More than happy to close this pr if you already have `OpenURLEffect` tests in progress else where.

zmian:
Closing for now and can reopen if necessary.

PR #59, opened 2023-03-06
Fire-and-forget should not participate in cooperative cancellation.

mbrandonw:


Tests/DependenciesTests/FireAndForgetTests.swift
```diff
@@ -16,6 +17,23 @@ final class FireAndForgetTests: XCTestCase {
     XCTAssertEqual(value, true)
   }
 
+  func testTestContext_Cancellation() async throws {
```
mbrandonw, line 20: this test failed without the change.

Tests/DependenciesTests/FireAndForgetTests.swift
```diff
@@ -36,21 +54,23 @@ final class FireAndForgetTests: XCTestCase {
     }
   }
 
-  func testLiveContext_DependencyAccess() async {
-    await withDependencies {
-      $0.context = .live
-      $0.date.now = Date(timeIntervalSince1970: 1_234_567_890)
-    } operation: {
-      let date = ActorIsolated<Date?>(nil)
+  #if !os(Linux)
```
mbrandonw, line 57: I'm able to reproduce this test hang locally but couldn't figure it out easily so just punting on it for now. maybe a linux concurrency bug :/

stephencelis (approved):

stephencelis (approved):

PR #60, opened 2023-03-06
Support environment variables overrides for context.

mbrandonw:
This makes it possible to override the context used for dependencies via an environment variable, which can be useful for UI integration tests. I also added some integration tests to this project to verify this works.

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -220,8 +220,6 @@ struct CurrentDependency {
   var line: UInt?
 }
 
-private let contextEnvVarName = "SWIFT_DEPENDENCIES_CONTEXT"
-
 private let defaultContext: DependencyContext = {
   let environment = ProcessInfo.processInfo.environment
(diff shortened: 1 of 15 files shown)


Sources/Dependencies/DependencyValues.swift
```diff
@@ -221,12 +221,33 @@ struct CurrentDependency {
...
-  } else if _XCTIsTesting {
+  case "test":
     return .test
-  } else {
-    return .live
+  default:
+    runtimeWarn("""
+      An environment value for SWIFT_DEPENDENCIES_CONTEXT was provided but did not match "live",
+      "preview", or "test".
+      """)
```
stephencelis, line 249:
```suggestion
    runtimeWarn(
      """
      An environment value for SWIFT_DEPENDENCIES_CONTEXT was provided but did not match "live",
      "preview", or "test".
      
          SWIFT_DEPENDENCIES_CONTEXT = \(value.debugDescription)
      """
    )
```

PR #68, opened 2023-03-21
constrain TestDependencyKey associatedtype to Sendable

adriansergheev:
This seems to be fixed: https://github.com/apple/swift/issues/60649
running swiftlang-5.7.2.135.5 clang-1400.0.29.51

### DIFF
--- a/Sources/Dependencies/DependencyKey.swift
+++ b/Sources/Dependencies/DependencyKey.swift
@@ -118,7 +118,7 @@ public protocol DependencyKey: TestDependencyKey {
 /// return a default value suitable for Xcode previews, or the ``testValue``, if left unimplemented.
 ///
 /// See ``DependencyKey`` to define a static, default value for the live application.
-public protocol TestDependencyKey: Sendable where Value: Sendable {
+public protocol TestDependencyKey {
   /// The associated type representing the type of the dependency key's value.
 	associatedtype Value = Self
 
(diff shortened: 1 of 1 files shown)


Sources/Dependencies/DependencyKey.swift
```diff
@@ -118,12 +118,9 @@ public protocol DependencyKey: TestDependencyKey {
 /// return a default value suitable for Xcode previews, or the ``testValue``, if left unimplemented.
 ///
 /// See ``DependencyKey`` to define a static, default value for the live application.
-public protocol TestDependencyKey {
+public protocol TestDependencyKey: Sendable where Value: Sendable {
```
stephencelis, line 121: We don't actually need the key type to conform, just the value, so we can move that constraint to the associated type.
```suggestion
public protocol TestDependencyKey {
```

Sources/Dependencies/DependencyKey.swift
```diff
@@ -118,12 +118,9 @@ public protocol DependencyKey: TestDependencyKey {
...
   /// The associated type representing the type of the dependency key's value.
-  associatedtype Value = Self
-
-  // NB: This associated type should be constrained to `Sendable` when this bug is fixed:
-  //     https://github.com/apple/swift/issues/60649
+	associatedtype Value = Self
```
stephencelis, line 123: Looks like the fix landed in 5.7.1: https://github.com/apple/swift/blob/swift-5.7.1-RELEASE/lib/Sema/CSSimplify.cpp#L3862-L3868

```suggestion
  #if swift(>=5.7.1)
  	associatedtype Value: Sendable = Self
  #else
    // NB: Can't constrain to `Sendable` on earlier Swift versions due to this bug:
    //     https://github.com/apple/swift/issues/60649
    associatedtype Value = Self
  #endif
```

stephencelis (approved):
Thanks! Just a couple changes to make and I'll merge!

PR #69, opened 2023-03-23
Add `UUID.init(_ intValue:)` helper

stephencelis:


acosmicflamingo:
Oh clever! I memorized 8-4-4-4-12 for creating UUIDs but looks like this will be helpful :)

Sources/Dependencies/DependencyValues/UUID.swift
```diff
@@ -104,6 +104,12 @@ public struct UUIDGenerator: Sendable {
   }
 }
 
+extension UUID {
+  public init(_ intValue: Int) {
```
tgrapperon, line 108: Is it working if we pass -1?

mbrandonw (approved):

Sources/Dependencies/DependencyValues/UUID.swift
stephencelis: > Is it working as expected if we pass -1?

I think we'd consider `-1` "undefined" and leave it at that.

> Otherwise, we could restrict to UInt or take the absolute value.

We prefer `Int` to `UInt` as Swift tends to do (using it for `Array.count`, `Array.index`, etc.).

We're open to improvements if folks wanna PR, but seeing as this is primarily a test helper, we're not too worried about it covering edge cases.

PR #73, opened 2023-04-01
Import Foundation in RuntimeWarnings and update test expectation methods

amonshiz:
New warning with Xcode 14.3 about not having access to Swift.String's CVarArgs conformance without the Foundation import. And using async specific expectation waiting method that is now a warning in Xcode 14.3

### DIFF
--- a/Sources/Dependencies/Internal/RuntimeWarnings.swift
+++ b/Sources/Dependencies/Internal/RuntimeWarnings.swift
@@ -35,6 +35,7 @@ func runtimeWarn(
 #if DEBUG
   #if canImport(os)
     import os
+    import Foundation
 
     // NB: Xcode runtime warnings offer a much better experience than traditional assertions and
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hi @amonshiz, unfortunately we can't use the new fancy `await self.fulfillment` stuff because GitHub CI [still doesn't have access to Xcode 14.3](https://github.com/actions/runner-images/issues/6426#issuecomment-1479974889). Were those changes made due to Xcode warnings? If so, we may need revert that and just accept the warnings in the test target for the time being.

mbrandonw:
@amonshiz Actually you can also silence those warnings by wrapping `self.wait` in a sync closure and evaluating immediately:

```swift
_ = { self.wait(for: [expectation], timeout: 1) }()
```

Can you give that a shot?

amonshiz:
Thank you! I'll give those a try. 

But yes, Xcode 14.3 has warnings for using the non-Concurrency form in a concurrent context.

Issue #77, opened 2023-04-10
When on `live` context with no `liveValue` provided, the `testValue` fallback isn't cached


oronbz:
### Description

As description suggests, `testValue` fallback isn't cached on `live` context.

lines 337...341 on `DependencyValues.swift`

```Swift
        // should we add this: self.cached[cacheKey] = AnySendable(Key.testValue)
        return Key.testValue // <--- no cache
      }

      self.cached[cacheKey] = AnySendable(value)
      return value

```
This issue causes eager memory release of the acquired dependency, since `@Dependency(\.dep) var dep` isn't strong referencing the dependency, only the cache does.

Why should we want this?
Sometimes when separating `TestDependencyKey` in one module and `DependencyKey` in another, and there's an example app on the `TestDependencyKey` module, and it's cumbersome to create a `liveValue` again for each of them.

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-dependencies/issues) or [discussion](https://github.com/pointfreeco/swift-dependencies/discussions).

### Expected behavior

I expect the `testValue` in this case to be held to cache as a strong reference to it.

### Actual behavior

When on `live` context with no `liveValue` provided, the `testValue` fallback isn't cached

### Dependencies version information

0.2.0

### Destination operating system

iOS 16.4

### Xcode version information

Xcode 14.3

stephencelis:
@oronbz Thanks for diving into this! It definitely looks like a good change, but we probably need some time to process all the implications. One thing that would help, if you have the time, is a failing test case that shows the problem. This would give us a jumping-off point to explore some edge cases and work towards a solution that can get merged. Wanna give that a shot?

oronbz:
Is there any chance that this can be revisited?

rapidlugo:
@oronbz Hey, thanks for sharing your findings. Curious if you found that making this cache change had any effect on the lifecycle of objects in an XCTest environment - anything weird crop up when using the updated cached values?

oronbz:
[...]

PR #78, opened 2023-04-10
Cache values when accessing test dependencies from live context

oronbz:
#77

### DIFF
--- a/Tests/DependenciesTests/DependencyValuesTests.swift
+++ b/Tests/DependenciesTests/DependencyValuesTests.swift
@@ -179,8 +179,8 @@ final class DependencyValuesTests: XCTestCase {
             reuseClient.setCount(-42)
             XCTAssertEqual(
               reuseClient.count(),
-              0,
-              "Don't cache dependency when using a test value in a live context"
+              -42,
+              "Dependency should cache when using a test value in a live context"
             )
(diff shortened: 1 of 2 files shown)


Tests/DependenciesTests/DependencyValuesTests.swift
```diff
@@ -185,22 +185,17 @@ final class DependencyValuesTests: XCTestCase {
...
-                """
-                @Dependency(\\.reuseClient)" has no live implementation, but was accessed from a live \
-                context.
-                """
-              )
-            }
-          #endif
-          XCTAssertEqual(reuseClient.count(), 0)
-        }
+        #if DEBUG
```
oronbz, line 188: The modifications here are to adjust to the fact the failure only happens once

oronbz:
@stephencelis @mbrandonw , hope it's ok mentioning you here. any chance you can look into it? As our project switch more and more dependencies to swift-dependencies this become a bigger of an issue for us.

stephencelis (approved):
@oronbz Sorry for the delay. We've dug a bit more deeply into the problem and agree that we should be caching in this case. Thanks again for the PR!

oronbz:
Lovely. Thanks @stephencelis

PR #80, opened 2023-04-12
Revert "remove #if DEBUG guard"

### LINKED ISSUE #79: Linux build with `--static-swift-stdlib` fails
### Description

Building (the swift-dependencies package itself or with the package as a dependency) fails on Linux when using `--static-swift-stdlib`.

I should be able to work around the lack of static linking of the stdlib by building without and copying all the libs like we used to in the past but if there's a simple fix to support this flag that'd be great.

### Checklist

### Expected behavior

```
docker run --rm -v "$PWD":/host -w /host swift:5.8-focal swift build --static-swift-stdlib
```

should succeed, just like

```
docker run --rm -v "$PWD":/host -w /host swift:5.8-focal swift build
```

does.

### Actual behavior

Fails to compile with error:

```
/host/Sources/Dependencies/DependencyValues.swift:353:10: error: no such module 'XCTest'
  import XCTest
         ^
```

### Steps to reproduce

Steps to reproduce:

```
❯ git clone https://github.com/pointfreeco/swift-dependencies && cd swift-dependencies
❯ git rev-parse @
0f34848e741315119703b1990e5c962929ed5dff
❯ docker run --rm -v "$PWD":/host -w /host swift:5.8-focal swift build --static-swift-stdlib
Fetching https://github.com/apple/swift-argument-parser
Fetching https://github.com/pointfreeco/xctest-dynamic-overlay
...
[96/122] Emitting module Dependencies
/host/Sources/Dependencies/DependencyValues.swift:353:10: error: no such module 'XCTest'
  import XCTest
         ^
[97/139] Compiling Benchmark BenchmarkCommand.swift
...
```

0f34848e741315119703b1990e5c962929ed5dff (current `main`)

Linux

### Xcode version information

_No response_

```shell
Swift version 5.8 (swift-5.8-RELEASE)
Target: aarch64-unknown-linux-gnu
```

finestructure:
This reverts commit cfdddbc569638fceb501ceca81ef6a23e490c311.

Fixes #79 

This isn't a complete fix in the sense that

```
docker run --rm -v "$PWD":/host -w /host swift:5.8-focal swift build --static-swift-stdlib
```

will work – but it'll allow _release_ builds to work, which is the context in which the build failure is actually relevant:

```
docker run --rm -v "$PWD":/host -w /host swift:5.8-focal swift build -c release --static-swift-stdlib
```

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -95,7 +95,9 @@ public struct DependencyValues: Sendable {
   /// provide access only to default values. Instead, you rely on the dependency values' instance
   /// that the library manages for you when you use the ``Dependency`` property wrapper.
   public init() {
-    _ = setUpTestObservers
+    #if DEBUG
+      _ = setUpTestObservers
+    #endif
   }
 
   /// Accesses the dependency value associated with a custom key.
@@ -273,7 +275,7 @@ private final class CachedValues: @unchecked Sendable {
   ) -> Key.Value where Key.Value: Sendable {
     self.lock.lock()
     defer { self.lock.unlock() }
-
+ 
     let cacheKey = CacheKey(id: ObjectIdentifier(key), context: context)
     guard let base = self.cached[cacheKey]?.base, let value = base as? Key.Value
     else {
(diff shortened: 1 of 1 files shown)


Sources/Dependencies/DependencyValues.swift
```diff
@@ -349,60 +351,62 @@ private final class CachedValues: @unchecked Sendable {
 
 // NB: We cannot statically link/load XCTest on Apple platforms, so we dynamically load things
 //     instead and we limit this to debug builds to avoid App Store binary rejection.
```
finestructure, line 353: Note that commit cfdddbc569638fceb501ceca81ef6a23e490c311 didn't update this comment here, causing code/comment drift.

Sources/Dependencies/DependencyValues.swift
```diff
@@ -95,7 +95,9 @@ public struct DependencyValues: Sendable {
   /// provide access only to default values. Instead, you rely on the dependency values' instance
   /// that the library manages for you when you use the ``Dependency`` property wrapper.
   public init() {
-    _ = setUpTestObservers
+    #if DEBUG
+      _ = setUpTestObservers
+    #endif
```
finestructure, line 100: I guess this one wouldn't need to be rolled back to fix the Linux release build but I started by simply reverting the whole patch for now.

Sources/Dependencies/DependencyValues.swift
```diff
@@ -349,60 +351,62 @@ private final class CachedValues: @unchecked Sendable {
 
 // NB: We cannot statically link/load XCTest on Apple platforms, so we dynamically load things
 //     instead and we limit this to debug builds to avoid App Store binary rejection.
-#if !canImport(ObjectiveC)
-  import XCTest
-#endif
+#if DEBUG
```
finestructure, line 354: I'm not sure what the reason was to remove the `DEBUG` guard but if it needs to be removed, perhaps a better solution would be something like

```
#if DEBUG || !os(Linux)
```

or maybe simply

```
#if !os(Linux)
```

?

I'm not sure what the block does exactly so it's hard to judge under what conditions it needs to be compiled in and what its absence means, functionality-wise.

mbrandonw: Hi @finestructure, we are wanting to remove the `#if DEBUG` because it is a bummer to have different behavior for debug vs release builds in tests. It basically means you cannot run your test code in release mode if it uses this library. The same is true of our xctest-dynamic-overlay library and we are hoping to remove the `#if DEBUG`s from that too.

It seems like we can instead wrap the entire testing logic in a `#if canImport(XCTest)` since that will be `false` when built with `--static-stdlib`, and it’s already not possible to run swift test `--static-stdlib` so we aren’t losing anything there.

What do you think of making that change instead? Or if you don't have time we can take care of it.

finestructure: I see, that makes sense! Let me update that real quick but if it's still not quite right please feel free to mess with the PR!

mbrandonw: Great, I just pushed some fixes and I added `swift build -static-stdlib` to CI so that we can catch these things in the future.

mbrandonw:
Hm, getting some failures with the new [CI step](https://github.com/pointfreeco/swift-dependencies/pull/80/commits/fb359d26d9a2c33e451b4090e8f51412190964d1) I added: https://github.com/pointfreeco/swift-dependencies/actions/runs/4681143371/jobs/8293354818?pr=80

I don't have time to look at this right now, but if you have some ideas let me know, and otherwise will take a look a bit later.

Makefile
```diff
@@ -35,6 +35,10 @@ test-linux:
 		swift:5.7-focal \
 		bash -c 'apt-get update && apt-get -y install make && make test-swift'
 
+build-for-static-stdlib:
+	@swift build -c debug -Xswiftc -static-stdlib
+	@swift build -c release -Xswiftc -static-stdlib
```
finestructure, line 40: note that I've been using `swift build --static-swift-stdlib` (double `-`) - not sure if that's the same as `-Xswiftc -static-stdlib`

Makefile
```diff
@@ -36,8 +36,8 @@ test-linux:
 		bash -c 'apt-get update && apt-get -y install make && make test-swift'
 
 build-for-static-stdlib:
-	@swift build -c debug -Xswiftc -static-stdlib
-	@swift build -c release -Xswiftc -static-stdlib
+	@swift build -c debug --static-swift-stdlib
+	@swift build -c release --static-swift-stdlib
```
mbrandonw, line 40: ahhh good catch, thanks!

finestructure:
Ok, I think this is due to differences in the 5.7 / 5.8 base image. CI fails with

```
/usr/bin/ld.gold: error: cannot find -lcurl
```

That works for me locally with `swift:5.8-focal`, which might not need `curl` anymore. Could either install `libcurl-dev` in the image or run it with 5.8.

finestructure:
Let me know if you prefer to have this run on 5.7 for now and I can fix that up tomorrow.

mbrandonw:
I think 5.8 is fine, but I'll discuss with @stephencelis. At the very least if this next CI run succeeds we can merge and get a release out to unblock people running into this.

Makefile
```diff
@@ -55,12 +55,12 @@ build-for-static-stdlib-docker:
 	@docker run \
 		-v "$(PWD):$(PWD)" \
 		-w "$(PWD)" \
-		swift:5.7-focal \
+		swift:5.8-focal \
```
mbrandonw: the `build-for-static-stdlib` command is what is run in CI, so we gotta update the CI steps to run in 5.8.

stephencelis:
@finestructure I'd push to your branch but don't have access.

I think you can add [this action](https://github.com/marketplace/actions/setup-swift) to the ci.yml, though:

```diff
     steps:
       - uses: actions/checkout@v3
+      - uses: swift-actions/setup-swift@v1
+        with:
+          swift-version: '5.8.0'
       - name: Build for static-stdlib
         run: make build-for-static-stdlib
```

finestructure:
swift-action isn't on 5.8 yet (last I checked, there's a PR open). We've forked it for our own use and added 5.8. I can point it to ours temporarily if you'd like?

I'm not at my computer right now but I'm happy to fix that all up in the morning!

I'll see if I can run the CI tasks in my fork or something, because right now each run needs approval I think.

finestructure:
> @finestructure I'd push to your branch but don't have access.

That’s odd, the “Allow edits by maintainers” checkbox is ticked 🤔

.github/workflows/ci.yml
```diff
@@ -49,3 +49,18 @@ jobs:
...
+    strategy:
+      matrix:
+        os: [ubuntu-20.04]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: swift-actions/setup-swift@v1
+        with:
+          swift-version: '5.8.0'
+      - name: Install dependencies
+        run: apt-get install -y libcurl4-openssl-dev
```
finestructure: Ok, the reason why this worked for me locally was that `swift:5.8-focal` comes with the `curl` dev library:

```
$ docker run --rm -it swift:5.8-focal
root@docker-desktop:/host# apt list --installed | grep curl
libcurl3-gnutls/now 7.68.0-1ubuntu2.16 arm64 [installed,local]
libcurl4-openssl-dev/now 7.68.0-1ubuntu2.18 arm64 [installed,local]
libcurl4/now 7.68.0-1ubuntu2.18 arm64 [installed,local]
```

`swift-actions/setup-swift` doesn't seem to be installing this.

I also didn't realise that `static-stdlib` isn't running `make build-for-static-stdlib-docker`. It be possible to do that via docker-in-docker (we do that on Gitlab) but it's probably not worth fiddling with that.

CI passes now: https://github.com/finestructure/swift-dependencies/actions/runs/4686542179

.github/workflows/ci.yml
```diff
@@ -49,3 +49,18 @@ jobs:
...
+    strategy:
+      matrix:
+        os: [ubuntu-20.04]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: swift-actions/setup-swift@v1
+        with:
+          swift-version: '5.8.0'
+      - name: Install dependencies
+        run: sudo apt-get install -y libcurl4-openssl-dev
```
finestructure, line 63: Ok, the reason why this worked for me locally was that `swift:5.8-focal` comes with the `curl` dev library:

```
$ docker run --rm -it swift:5.8-focal
root@docker-desktop:/host# apt list --installed | grep curl
libcurl3-gnutls/now 7.68.0-1ubuntu2.16 arm64 [installed,local]
libcurl4-openssl-dev/now 7.68.0-1ubuntu2.18 arm64 [installed,local]
libcurl4/now 7.68.0-1ubuntu2.18 arm64 [installed,local]
```

`swift-actions/setup-swift` doesn't seem to be installing this.

I also didn't realise that `static-stdlib` isn't running `make build-for-static-stdlib-docker`. It be possible to do that via docker-in-docker (we do that on Gitlab) but it's probably not worth fiddling with that.

CI passes now: https://github.com/finestructure/swift-dependencies/actions/runs/4686542179

PR #82, opened 2023-04-14
Simplify static-stdlib CI job

finestructure:
Proposing a small change to the static stdlib CI job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
   static-stdlib:
     strategy:
       matrix:
-        image: [swift:5.8-focal]
+        image: ['swift:5.8-focal']
     runs-on: ubuntu-latest
     container:
       image: ${{ matrix.image }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,11 +53,13 @@ jobs:
   static-stdlib:
     strategy:
       matrix:
-        image: ['swift:5.8-focal']
+        image: ['swift:5.7-focal', 'swift:5.8-focal']
     runs-on: ubuntu-latest
     container:
       image: ${{ matrix.image }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -53,14 +53,13 @@ jobs:
...
+        image: ['swift:5.7-focal', 'swift:5.8-focal']
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ matrix.image }}
     steps:
-      - uses: swift-actions/setup-swift@v1
-        with:
-          swift-version: '5.8.0'
-      - name: Install dependencies
-        run: sudo apt-get install -y libcurl4-openssl-dev
```
finestructure, line 63: The advantage is that it avoids the `apt-get install` and allows easy extension to test it on 5.7 as well.

The downside is that the image doesn't come with `make`, so the `swift build` commands have to be inlined.

NB: this mechanism could also be used to extend the `ubuntu-tests` to cover other swift versions.

stephencelis:
@finestructure I'm not super familiar with the different "flavors" of Swift images, but is "focal" likely to be used outside Docker builds running on M1/2 machines? My main thought is we should maybe stick with the most generic `swift` image in CI, but I admit I don't know the exact process that goes into selecting a specific image.

finestructure:
I'd say `focal` (ubuntu 20.04 LTS) is pretty generic. `bionic` (18.04 LTS), `focal` (20.04 LTS), and `jammy` (22.04 LTS) are the default ubuntu images / distros you get from https://www.swift.org/download/. I'm pretty sure `focal` is the most commonly used one, but `jammy` would work equally well.

The main point here is that it's an official swift.org image and you don't have to worry about baking in compiler dependencies.

If you'd rather stick with the current version where that's explicit feel free to just close the PR! It's merely a suggestion :)

stephencelis (approved):
Ah, TIL! Thanks Sven! I think this is good to take then :smile:

PR #84, opened 2023-04-14
Reintroduce withMainSerialExecutor

kabiroberai:
This PR reintroduces `withMainSerialExecutor` using `dlsym` rather than a C module, skirting around the issues highlighted in #83.

Diff between the old and new approaches: https://github.com/pointfreeco/swift-dependencies/compare/cd4edcf...4babbda

### DIFF
--- a/Sources/Dependencies/ConcurrencySupport/MainSerialExecutor.swift
+++ b/Sources/Dependencies/ConcurrencySupport/MainSerialExecutor.swift
@@ -16,15 +16,6 @@ import Foundation
 private typealias Orig = @convention(thin) (UnownedJob) -> Void
 private typealias Hook = @convention(thin) (UnownedJob, Orig) -> Void
 private var swift_task_enqueueGlobal_hook: UnsafeMutablePointer<Hook>? = {
-  var info = Dl_info()
-  guard
-    withUnsafePointer(to: TaskPriority.self, {
-      $0.withMemoryRebound(to: UnsafeRawPointer.self, capacity: 1) {
-        dladdr($0.pointee, &info)
-      }
-    }) != 0,
-    let handle = dlopen(info.dli_fname, RTLD_LAZY),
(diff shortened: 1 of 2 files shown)


Sources/Dependencies/ConcurrencySupport/MainSerialExecutor.swift
```diff
@@ -0,0 +1,30 @@
...
+  return try await operation()
+}
+
+// here be dragons
+private typealias Orig = @convention(thin) (UnownedJob) -> Void
+private typealias Hook = @convention(thin) (UnownedJob, Orig) -> Void
+private var swift_task_enqueueGlobal_hook: UnsafeMutablePointer<Hook>? = {
+  var info = Dl_info()
+  guard
+    withUnsafePointer(to: TaskPriority.self, {
```
kabiroberai, line 21: using this to obtain a handle to `libswift_Concurrency.{dylib,so}` without hardcoding a path — open to suggestions on better ways to do this. Ideally the image path would be resolved at launch time by the linker (which can be accelerated by dyld closures etc) but since this API should really only be used in tests I'm begrudgingly okay with the `dladdr` approach.

kabiroberai:
ugh looks like `dladdr` [requires](https://forums.swift.org/t/dladdr-and-the-clang-importer/26379) `#define _GNU_SOURCE` on Linux. We can call it through a C target but… you see where I'm going with this 😶

kabiroberai:
think we're good now! switched to searching the global namespace for the symbol, which isn't ideal performance-wise but it's amortized to be one-time + lazy and it's just for tests anyway

stephencelis:
@kabiroberai This is great, thanks! We plan on merging it but are thinking through where in the dependency chain we want to include it.

stephencelis:
Gonna merge as is for now, thanks again!

PR #86, opened 2023-04-18
Introduce Scheduler.dependencies()

kabiroberai:
This API enables the creation of schedulers that propagate dependencies across `schedule` calls. For motivation and further discussion, see #85.

### DIFF
--- a/Sources/Dependencies/DependencyValues/MainQueue.swift
+++ b/Sources/Dependencies/DependencyValues/MainQueue.swift
@@ -6,8 +6,9 @@
     ///
     /// Introduce controllable timing to your features by using the ``Dependency`` property wrapper
     /// with a key path to this property. The wrapped value is a Combine scheduler with the time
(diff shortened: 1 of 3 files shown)


Sources/Dependencies/DependencyValues/MainQueue.swift
```diff
@@ -53,7 +54,7 @@
     }
 
     private enum MainQueueKey: DependencyKey {
-      static let liveValue = AnySchedulerOf<DispatchQueue>.main
+      static let liveValue = DispatchQueue.main.dependencies()
       static let testValue = AnySchedulerOf<DispatchQueue>
         .unimplemented(#"@Dependency(\.mainQueue)"#)
```
kabiroberai, line 59: I was worried that users would explicitly need to use `dependencies()` in the test value, but then I realized that test schedulers already forward dependencies since they invoke the closure immediately (inheriting task locals). Though I'm not sure if this applies to the delay/interval APIs; dunno if we need to add explicit tests for those.

PR #97, opened 2023-05-31
Add test/docs around static `@Dependency` usage

stephencelis:
That is: don't.

Because statics are lazy, they will cache the current dependency values when first accessed, leading to unexpected behavior.


acosmicflamingo:
I wonder who found this out the hard way 😂

mbrandonw (approved):

PR #118, opened 2023-09-18
Make merging public

yonaskolb:
This makes merging of DependencyValues public, for outside usage


mbrandonw:
Hi @yonaskolb, can you explain a bit why you need access to this functionality?

stephencelis:
@yonaskolb That sounds interesting but I think we would need more info about how you're using it all. I don't know if you have time to write up a more in depth explanation, _e.g._ how the `merge` method could be documented and motivated for use outside the library, but it would be helpful for us if we want to motivate publicizing such an API.

stephencelis:
@yonaskolb I'm going to close this for now, but if you have something to share and want to reopen again in the future, let us know!

PR #126, opened 2023-10-10
Add public method to clear dependency cache

DylanBettermannDD:
We need a way to clear the dependency cache outside of an XCTest context.

Summary of changes:
1. Adds a public `DependencyValues.clearCache()` function which is an proxy for `cachedValues.cached = [:]`.
2. Converts the existing callsites over to the new function.
3. Changes `DependencyValues.cachedValues` from `fileprivate` to `private` given the new function.

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -371,7 +371,7 @@ private final class CachedValues: @unchecked Sendable {
               .takeUnretainedValue()
           else { return }
           let testCaseWillStartBlock: @convention(block) (AnyObject) -> Void = { _ in
-            DependencyValues._current.cachedValues.cached = [:]
+            DependencyValues._current.clearCache()
           }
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @DylanBettermannDD, can you provide some context for why you need this?

This is a pretty big change, and quite a huge departure from one of the library's core tenants: structured programming. One of the powers of the library is that one can only change dependencies in a lexical scope via `withDependencies`. If we provide a tool to clear the cache then we break that rule and it makes it harder to understand how dependencies change in an application.

One way to clear the cache while still remaining in the realms of structured programming is to reset the entire dependency values inside `withDependencies`:

```swift
withDependencies {
  $0 = DependencyValues()
} operation: {
  // cache is effectively cleared here
}
```

DylanBettermannDD:
Discussed in a different venue and I'm going to explore other options.

PR #130, opened 2023-10-11
Add docs about test leakage.

mbrandonw:
This came up in #129, and it has come up before, so thought it would be nice to have some docs on it.

### DIFF
--- a/Sources/Dependencies/Documentation.docc/Articles/Testing.md
+++ b/Sources/Dependencies/Documentation.docc/Articles/Testing.md
@@ -168,4 +168,39 @@ transitively get access to it through the app itself. In Xcode, go to "Build Pha
 "Dependencies" from the "Link Binary With Libraries" section. When using SwiftPM, remove the
 "Dependencies" entry from the `testTarget`'s' `dependencies` array in `Package.swift`.
 
+### Test case leakage
+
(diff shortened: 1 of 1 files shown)


Sources/Dependencies/Documentation.docc/Articles/Testing.md
```diff
@@ -168,4 +168,39 @@ transitively get access to it through the app itself. In Xcode, go to "Build Pha
 "Dependencies" from the "Link Binary With Libraries" section. When using SwiftPM, remove the
 "Dependencies" entry from the `testTarget`'s' `dependencies` array in `Package.swift`.
 
+### Test case leakage
+
+Sometimes it is possible to have tests that pass successfully when run in isolation, but somehow
```
dafurman, line 173:
```suggestion
Sometimes it is possible to have tests that pass successfully when run in isolation, but somehow fail
```
Just noticed the missing word here. Anyways, thanks for adding docs on this! Much appreciated as these will be handy to reference to my team when this situation comes up.

mbrandonw: good catch, thanks!

stephencelis (approved):

PR #145, opened 2023-11-22
Update README.md

kgrigsby59:
Correct SPI documentation link.


stephencelis (approved):
The other one does apparently work for me, but this does appear to be their canonical URL, thanks!

PR #158, opened 2023-12-09
Fix fatalerror default

### LINKED ISSUE #148: @DependencyEndpoint & Never returning closure
### Description

When applying `@DependencyEndpoint` to a function which return a non-optional type one must provide a closure. While this moved some runtime checks to compile time checks it's not always possible to provide a sensible value or even a value at all.
One can avoid providing a value using `fatalError()` but the macro then emits invalid swift code.

### Checklist

### Expected behavior

The macro should emit valid swift code even when using `fatalError()` but I think there should be an option to specify the behaviour of the macro so one can skip writing the fatalError entirely

### Actual behavior

_No response_

### Steps to reproduce

```
@DependencyEndpoint
public var fetchCategoryIds: () async throws -> Set<Article.Category.ID> = { fatalError() }
  private var _fetchCategoryIds: () async throws -> Set<Article.Category.ID> = {
      XCTestDynamicOverlay.XCTFail("Unimplemented: 'fetchCategoryIds'")
      return fatalError()
  }
{
    @storageRestrictions(initializes: _fetchCategoryIds)
    init(initialValue) {
        _fetchCategoryIds = initialValue
    }
    get {
        _fetchCategoryIds
    }
    set {
        _fetchCategoryIds = newValue
    }
}
```

1.1.1

_No response_

### Xcode version information

_No response_

_No response_

mbrandonw:
Fixes #148.

We don't recommend people use `fatalError`'s in their defaults, but we might as well generate valid Swift code if someone has a good reason to.

### DIFF
--- a/Sources/DependenciesMacros/Macros.swift
+++ b/Sources/DependenciesMacros/Macros.swift
@@ -164,3 +164,10 @@ public struct Unimplemented: Error {
     self.endpoint = endpoint
   }
 }
+
+struct Blah {
+  @DependencyEndpoint
+  public var foo: () -> String = { fatalError() }
+  @DependencyEndpoint
+  public var bar: () -> String = { fatalError("Goodbye") }
+}
--- a/Sources/DependenciesMacros/Macros.swift
+++ b/Sources/DependenciesMacros/Macros.swift
@@ -164,10 +164,3 @@ public struct Unimplemented: Error {
     self.endpoint = endpoint
   }
 }
-
-struct Blah {
-  @DependencyEndpoint
-  public var foo: () -> String = { fatalError() }
-  @DependencyEndpoint
-  public var bar: () -> String = { fatalError("Goodbye") }
-}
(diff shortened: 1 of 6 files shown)


Sources/DependenciesMacros/Macros.swift
```diff
@@ -164,3 +164,10 @@ public struct Unimplemented: Error {
     self.endpoint = endpoint
   }
 }
+
+struct Blah {
+  @DependencyEndpoint
+  public var foo: () -> String = { fatalError() }
+  @DependencyEndpoint
+  public var bar: () -> String = { fatalError("Goodbye") }
+}
```
stephencelis, line 173:
```suggestion
```

Tests/DependenciesMacrosPluginTests/DependencyClientMacroTests.swift
```diff
@@ -513,22 +513,6 @@ final class DependencyClientMacroTests: BaseTestCase {
...
-        init(
-          endpoint: @Sendable @escaping () -> Int
-        ) {
-          self.endpoint = endpoint
-        }
-
-        init() {
-        }
-      }
-      """
```
mbrandonw, line 531: this was failing for me and so I had remove it. but not really sure why it doesn't fail on main

stephencelis (approved):

PR #159, opened 2023-12-09
Fix autoclosure endpoints

mbrandonw:
Fixes #155.

PR #160, opened 2023-12-10
Support `@autoclosure` within arguments at `@DependencyEndpoint`

### LINKED ISSUE #155: `@DependencyClient` and `@autoclosure` parameters
### Description

This is similar to https://github.com/pointfreeco/swift-dependencies/issues/146. The code generated by `@DependencyClient` is incorrect for `@autoclosure` parameters and generate invalid swift code. it should explicitly refuse to generate code or if possible generate valid code.

### Checklist

- [ ] If possible, I've reproduced the issue using the `main` branch of this package.

### Expected behavior

```swift
@DependencyClient
public struct DemoClinet {
    public var execute: (_ closure: @autoclosure () -> String) -> Void
}
```

Should generate valid swift code

### Actual behavior

The code generated :

```swift
public func execute(closure p0: @autoclosure () -> String) -> Void {
    self.execute(p0)  // Error occurs: 'Add () to forward @autoclosure parameter'
}

private var _execute: (_ closure: @autoclosure () -> String) -> Void = { _ in
    XCTestDynamicOverlay.XCTFail("Unimplemented: 'execute'")
}
```

is not valid code: `Add () to forward @autoclosure parameter`

### Steps to reproduce

_No response_

1.1.2

_No response_

### Xcode version information

15.0.0

_No response_

xiii111:
Resolve #155 

## Cause
```swift
struct Foo {
  @DependencyEndpoint
  var bar: (_ a: Int) -> Void
}
```

In general, the code as shown can typically be transformed into the following without any issues. This is because the type of `a` that `func bar` receives matches the type of the argument `a` passed to `var bar`.

```swift
func bar(a p0: Int) -> Void {
  self.bar(p0)
}
```

This also applies to simple closures.

```swift
struct Foo {
  @DependencyEndpoint
  var bar: (_ a: () -> Int) -> Void
}

func bar(a p0: () -> Int) -> Void {
  self.bar(p0)
}
```

However, when considering closures marked with `@autoclosure`, upon closer examination of @autoclosure behavior, the type of `a` that `func bar` receives might not match the type of the argument `a` passed to `var bar`.

```swift
struct Foo {
  @DependencyEndpoint
  var bar: (_ a: () @autoclosure -> Int) -> Void
}

func bar(a p0: @autoclosure () -> Int) -> Void {
  self.bar(p0) // Error occurs: 'Add () to forward @autoclosure parameter'
}
```

## Solution
Simply adding `()` should resolve this bug.

```swift
struct Foo {
  @DependencyEndpoint
  var bar: (_ a: () @autoclosure -> Int) -> Void
}

func bar(a p0: @autoclosure () -> Int) -> Void {
  self.bar(p0()) // No Error occurs
}
```

## Concern
I might be overly concerned, but I feel there's a slight difference between directly invoking `var bar` and indirectly through `func bar`. When directly calling `var bar`, it receives `() -> String`. However, when calling `var bar` indirectly through `func bar`, it receives `() -> (() -> String)()`. I don't expect this difference to cause any change in behavior, but if there might be an issue, I'd appreciate some advice!

### DIFF
--- a/Sources/DependenciesMacrosPlugin/DependencyEndpointMacro.swift
+++ b/Sources/DependenciesMacrosPlugin/DependencyEndpointMacro.swift
@@ -162,14 +162,14 @@ public enum DependencyEndpointMacro: AccessorMacro, PeerMacro {
       }
       let appliedParameters =
         parameters
-        .map {
-          guard let typed = $0.type.as(AttributedTypeSyntax.self),
-            typed.specifier?.tokenKind == .keyword(.inout)
-          else { return false }
-          return true
-        }
         .enumerated()
-        .map { $1 ? "&p\($0)" : "p\($0)" }
+        .map {
+          $1.isInout
+          ? "&p\($0)"
+          : $1.isAutoclosure
+          ? "p\($0)()"
+          : "p\($0)"
+        }
         .joined(separator: ", ")
       decls.append(
         """
@@ -254,5 +254,28 @@ extension String {
       result = result.dropLast()
     }
     return String(result)
+  }
+}
+
+extension TupleTypeElementSyntax {
+  fileprivate var isAutoclosure: Bool {
+    self.type
+      .as(AttributedTypeSyntax.self)?
+      .attributes
+      .contains {
+        $0
+          .as(AttributeSyntax.self)?
+          .attributeName
+          .as(IdentifierTypeSyntax.self)?
+          .name
+          .tokenKind == .identifier("autoclosure")
+      } ?? false
+  }
+
+  fileprivate var isInout: Bool {
+    self.type
+      .as(AttributedTypeSyntax.self)?
+      .specifier?
+      .tokenKind == .keyword(.inout)
   }
 }
--- a/Tests/DependenciesMacrosPluginTests/DependencyEndpointMacroTests.swift
+++ b/Tests/DependenciesMacrosPluginTests/DependencyEndpointMacroTests.swift
@@ -796,4 +796,40 @@ final class DependencyEndpointMacroTests: BaseTestCase {
     """
     }
   }
+
+  func testAutoclosure() {
+    assertMacro {
+      """
+      struct Foo {
+        @DependencyEndpoint
+        var bar: (_ a: @autoclosure () -> Int, _ b: () -> Int, _ c: @autoclosure () -> Int) -> Void
+      }
+      """
+    } expansion: {
+      """
+      struct Foo {
+        var bar: (_ a: @autoclosure () -> Int, _ b: () -> Int, _ c: @autoclosure () -> Int) -> Void {
+          @storageRestrictions(initializes: _bar)
+          init(initialValue) {
+            _bar = initialValue
+          }
+          get {
+            _bar
+          }
+          set {
+            _bar = newValue
+          }
+        }
+
+        func bar(a p0: @autoclosure () -> Int, b p1: () -> Int, c p2: @autoclosure () -> Int) -> Void {
+          self.bar(p0(), p1, p2())
+        }
+
+        private var _bar: (_ a: @autoclosure () -> Int, _ b: () -> Int, _ c: @autoclosure () -> Int) -> Void = { _, _, _ in
+          XCTestDynamicOverlay.XCTFail("Unimplemented: 'bar'")
+        }
+      }
+      """
+    }
+  }
 }


Sources/DependenciesMacrosPlugin/DependencyEndpointMacro.swift
```diff
@@ -162,14 +162,14 @@ public enum DependencyEndpointMacro: AccessorMacro, PeerMacro {
...
         .map {
-          guard let typed = $0.type.as(AttributedTypeSyntax.self),
-            typed.specifier?.tokenKind == .keyword(.inout)
-          else { return false }
-          return true
+          $1.isInout
+          ? "&p\($0)"
+          : $1.isAutoclosure
+          ? "p\($0)()"
+          : "p\($0)"
```
xiii111, line 171: #159 style seems so cool that I had to try it myself. Originally, I was writing code like this 😅

```swift
let appliedParameters =
  parameters
  .enumerated()
  .map {
    if let typed = $1.type.as(AttributedTypeSyntax.self),
       typed.specifier?.tokenKind == .keyword(.inout) {
      return "&p\($0)"
    }
    
    if let typed = $1.type.as(AttributedTypeSyntax.self),
       typed.attributes.contains(where: {
         $0.as(AttributeSyntax.self)?.attributeName.as(IdentifierTypeSyntax.self)?.name.tokenKind == .identifier("autoclosure")
       }) {
      return "p\($0)()"
    }

    return "p\($0)"
  }
  .joined(separator: ", ")
```

Sources/DependenciesMacrosPlugin/DependencyEndpointMacro.swift
```diff
@@ -256,3 +256,26 @@ extension String {
...
+    self.type
+      .as(AttributedTypeSyntax.self)?
+      .attributes
+      .contains {
+        $0
+          .as(AttributeSyntax.self)?
+          .attributeName
+          .as(IdentifierTypeSyntax.self)?
+          .name
+          .tokenKind == .identifier("autoclosure")
```
xiii111, line 271: This code `.tokenKind` differs from #159. I'm thinking that `tokenKind` might be better than `text`, but I'll defer to your opinion.

Sources/DependenciesMacrosPlugin/DependencyEndpointMacro.swift
```diff
@@ -256,3 +256,26 @@ extension String {
     return String(result)
   }
 }
+
+extension TupleTypeElementSyntax {
+  fileprivate var isAutoclosure: Bool {
```
xiii111, line 261: To obtain this code, I used the Swift AST Explorer. When I inputted

```swift
var bar: (_ a: @autoclosure () -> Int) -> Void
```

it produced the following output.

スクリーンショット 2023-12-10 19 38 27

Tests/DependenciesMacrosPluginTests/DependencyEndpointMacroTests.swift
```diff
@@ -796,4 +796,40 @@ final class DependencyEndpointMacroTests: BaseTestCase {
     """
     }
   }
+
+  func testAutoclosure() {
+    assertMacro {
+      """
+      struct Foo {
+        @DependencyEndpoint
+        var bar: (_ a: @autoclosure () -> Int, _ b: () -> Int, _ c: @autoclosure () -> Int) -> Void
```
xiii111, line 805: I aligned the naming with #159, like `Foo` and `Bar`. The difference from #159 lies in using `b` as a closure. I intended to make it clear what differences occur based on whether `@autoclosure` is used or not.

Sources/DependenciesMacrosPlugin/DependencyEndpointMacro.swift
```diff
@@ -162,14 +162,14 @@ public enum DependencyEndpointMacro: AccessorMacro, PeerMacro {
...
+        .enumerated()
         .map {
-          guard let typed = $0.type.as(AttributedTypeSyntax.self),
-            typed.specifier?.tokenKind == .keyword(.inout)
-          else { return false }
-          return true
+          $1.isInout
+          ? "&p\($0)"
+          : $1.isAutoclosure
+          ? "p\($0)()"
```
xiii111, line 170: Only `()` is OK because `@autoclosure` can be used only when there is no parameters.
Also `inout` expression cannot be used on `@autoclosure` so we don't need to consider the situation when `$1.isInout && $1.isAutoclosure`.

mbrandonw (approved):
Thanks for taking the time to PR this!

mbrandonw (approved):
Great, thanks @xiii111!

xiii111:
@mbrandonw Thank you!

PR #164, opened 2023-12-15
Update docs for @DependencyClient about default values.

mbrandonw:


Sources/DependenciesMacrosPlugin/Support.swift
```diff
@@ -188,6 +189,14 @@ extension MacroExpansionContext {
         message: MacroExpansionErrorMessage(
           """
           Default value required for non-throwing closure '\(identifier)'
+
+          Defaults are required so that the macro can generate a default, "unimplemented" version \
+          of the dependency\(clientName.map { " via \($0)()"} ?? ""). The default value can be \
```
stephencelis, line 194:
```suggestion
          of the dependency\(clientName.map { " via '\($0)()'"} ?? ""). The default value can be \
```

stephencelis (approved):

PR #167, opened 2024-01-02
Clearer warning message

ZevEisenberg:
Old warning message:

>Do not use 'unimplemented' with '@DependencyClient'; it is a replacement and implements the same runtime functionality as 'unimplemented' at compile time

This was confusing because it was not clear what “it” referred to, nor what was a “replacement” for what.

New warning message:

>Do not use 'unimplemented' with '@DependencyEndpoint'; the '@DependencyEndpoint' macro provides an 'unimplemented' default value for you.

I also added a period at the end, but I can remove it if it’s not consistent with project style.

### DIFF
--- a/Sources/DependenciesMacrosPlugin/Support.swift
+++ b/Sources/DependenciesMacrosPlugin/Support.swift
@@ -96,8 +96,9 @@ extension InitializerClauseSyntax {
             node: self.value,
             message: MacroExpansionWarningMessage(
               """
-              Do not use 'unimplemented' with '@\(attribute.attributeName)'; it is a replacement and \
-              implements the same runtime functionality as 'unimplemented' at compile time
+              Do not use 'unimplemented' with '@\(attribute.attributeName)'; the \
+              '@\(attribute.attributeName)' macro already includes the behavior of \
+              'unimplemented'.
               """
             )
           )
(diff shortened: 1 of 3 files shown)


Sources/DependenciesMacrosPlugin/Support.swift
```diff
@@ -96,8 +96,8 @@ extension InitializerClauseSyntax {
             node: self.value,
             message: MacroExpansionWarningMessage(
               """
-              Do not use 'unimplemented' with '@\(attribute.attributeName)'; it is a replacement and \
-              implements the same runtime functionality as 'unimplemented' at compile time
+              Do not use 'unimplemented' with '@\(attribute.attributeName)'; the '@\(attribute.attributeName)' macro \
+              provides an 'unimplemented' default value for you.
```
stephencelis, line 100: Thoughts on a lil more workshopping?

```suggestion
              Do not use 'unimplemented' with '@\(attribute.attributeName)'; the \
              '@\(attribute.attributeName)' macro already includes the behavior of \
              'unimplemented'.
```

ZevEisenberg:
Updated with workshopped message.

stephencelis (approved):
Thanks!

PR #168, opened 2024-01-12
[MOD] modified version on the code snippet

x-0o0:
This is very minor modification.
I just found that the article guides to use old version (`0.1.0`)
I modified it to `1.0.0`


stephencelis (approved):
Thanks!

PR #169, opened 2024-01-16
Dependencies: dynamically link XCTest

compnerd:
Use delayed dynamic linking for registering the test observer. This is particularly important for Windows which does not support weak linking and weak symbols. Even in the case that `Dependencies` is imported due to use, we will not pull in the test observer until runtime, as a best effort. If the DependenciesTestObserver library is found in the search path, it will be loaded and the observer initialised. If we can't find the library, we will simply continue without the test observer.

### DIFF
--- a/Package@swift-5.9.swift
+++ b/Package@swift-5.9.swift
@@ -19,7 +19,7 @@ let package = Package(
     .library(
       name: "DependenciesMacros",
       targets: ["DependenciesMacros"]
-    )
+    ),
   ],
   dependencies: [
     .package(url: "https://github.com/apple/swift-syntax", from: "509.0.0"),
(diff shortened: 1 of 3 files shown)


compnerd:
Interesting, the tests pass with `--parallel`, but fail without. Am I mixing up some state?

compnerd:
Okay, preserving the existing behaviour on Darwin seems reasonable - the difference is the ObjC runtime is hooked dynamically already.  The isolated library approach applies to non-ObjC runtimes (Linux, Windows). This I think might be down to 2 failures on Linux (which seem like a bug in the test case).

compnerd:
Taking a different approach to isolate the XCTest dependency, but without the SPI so conceptually simpler but with more indirection. I think that this is a better shape overall. Seems to pass on Linux and macOS.

compnerd:
Ah, right, WASI does not yet do dynamic modules. What do you recommend here @stephencelis?

stephencelis:
@compnerd I haven't dug in too deeply, but can we fall back to explicit importing in this case?

compnerd:
Okay, sure, I can make that change for WASI.

stephencelis (approved):
Thanks!

PR #170, opened 2024-01-17
Tests: simplify WASI handling in DependencyValues test

compnerd:
The tests were duplicated rather than just isolating the behavioural difference between WASI and non-WASI targets. Simplify to avoid duplicating the entire test.


stephencelis (approved):
Thanks!

PR #172, opened 2024-01-22
Add support for `@Dependency(MyDependency.self)`

stephencelis:
Now that `@Environment` supports a shorthand for objects, we should do similar in our library.

This PR allows you to skip the step of extending `DependencyValues` with a key path and instead rely directly on the dependency key, which usually aligns with the value.

```diff
 extension APIClient: DependencyKey {
   var liveValue: Self { … }
 }

-extension DependencyValues {
-  var apiClient: APIClient {
-    get { self[APIClient.self] }
-    set { self[APIClient.self] = newValue }
-  }
-}
-
 class MyModel {
-  @Dependency(\.apiClient) var apiClient
+  @Dependency(APIClient.self) var apiClient
```

This new API is less discoverable and friendly to autocomplete, but may be appropriate for application code that refers to a singleton type, and can save that bit of boilerplate above.

### DIFF
--- a/Sources/Dependencies/DependencyKey.swift
+++ b/Sources/Dependencies/DependencyKey.swift
@@ -210,25 +210,30 @@ extension DependencyKey {
               \(typeName(Value.self))
           """
       )
-      let dependencyName =
+
+      let (argument, override) =
         DependencyValues.currentDependency.name
-        .map { "@Dependency(\\.\($0))" }
-        ?? "A dependency"
+        .map {
+          "\($0)" == "subscript(_:)"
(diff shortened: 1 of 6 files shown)


Sources/Dependencies/DependencyKey.swift
```diff
@@ -58,7 +58,7 @@
 /// implementation module.
 ///
 /// See the <doc:LivePreviewTest> article for more information.
-public protocol DependencyKey: TestDependencyKey {
+public protocol DependencyKey<Value>: TestDependencyKey {
```
stephencelis, line 61: Snuck this feature into the branch, but could extract to a separate PR...

mbrandonw (approved):

PR #196, opened 2024-03-15
Fix assigned static properties

rhysm94:
Ensures that the `@DependencyClient` macro doesn't attempt to do anything with static properties.
The macro bails out of looking a specific member property if the member is `static`. Wouldn't make sense to try and include them in the initialiser or add the `@DependencyEndpoint` to them, so I figured it was worth just bailing as quickly as possible on them.

I did notice that some of the other unit tests, the DependencyEndpoint tests, are failing. They were failing before I did any of this work though.

As discussed in the Slack, without this, assigned static properties in the form: `static var prop = SomeValue()` will crash the compiler, as they're treated the exact same as `var prop: SomeValue = SomeValue()`.

image


stephencelis (approved):
Thanks!

PR #197, opened 2024-03-22
Add safari dependency values

d-date:
Using `SFSafariViewController` in the context of SwiftUI does not work well around navigation. 
Therefore, using it in the context of UIKIt works better.  

In other platform, `EnvironmentValues().openURL` is used same as `@Dependency(\.openURL)` .

cf: https://sarunw.com/posts/sfsafariviewcontroller-in-swiftui/

### DIFF
--- a/Sources/Dependencies/DependencyValues/Safari.swift
+++ b/Sources/Dependencies/DependencyValues/Safari.swift
@@ -1,0 +1,88 @@
+#if canImport(SafariServices) && canImport(SwiftUI)
+import SafariServices
+import SwiftUI
+
+extension DependencyValues {
+  /// A dependency that opens a URL in SFSafariViewController.
+  ///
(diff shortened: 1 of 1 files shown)


stephencelis:
Thanks for the PR @d-date! You're packing your schedule quite a lot :smile:

Do you have a sample code showing the problem in UIKit using `openURL`? If there's a problem there I wonder if we should even update `openURL` itself.

d-date:
Hi @stephencelis and @mbrandonw,
Your talk and workshop are really awesome! I'll work with you as a contributor 😄 

I mean `@Dependency(\.openURL)` works well, but when somebody wants to use `SFSafariViewController`, they lost way without workaround. This PR save them.

You can check our PR and how we modified.
https://github.com/tryswift/trySwiftTokyoApp/pull/45

PR #198, opened 2024-03-23
Fix dependency value type in documentation

gibachan:
In the context of this explanation, using `UserDefaults` seems more suitable than `APIClient`.


stephencelis (approved):
Ha yes, thank you for catching this and taking the time to fix!

PR #203, opened 2024-04-02
Add NotificationCenter dependency value

gibachan:
We often use NotificationCenter to broadcast notifications in the app, so it would be useful to have it as a dependency.

### DIFF
--- a/Sources/Dependencies/DependencyValues/NotificationCenter.swift
+++ b/Sources/Dependencies/DependencyValues/NotificationCenter.swift
@@ -1,0 +1,29 @@
+import Foundation
+
+extension DependencyValues {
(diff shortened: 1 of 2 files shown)


stephencelis:
@gibachan Thanks for the PR! We've actually found that navigation center should be testable out of the box _without_ a dedicated dependency value, since you are free to post notifications to it yourself. Can you think of a case where overriding the center is required and where we might need a dependency value to override?

gibachan:
Ah, yeah, that makes sense.
I used to mock NotificationCenter to narrow down the scope for testing. That's why I thought we needed to override it, but now I see we don't really need it. Thank you for your comment!

PR #209, opened 2024-04-30
Add support for package access modifier to DependencyClient macro

ivanlisovyi:
## Context
Hey folks 👋  It seems like the `@DependencyClient` macro doesn't support the `package` access modifier when generating `init`'s, which makes it harder to split the interface and implementation into two separate targets if the interface uses `package` access. This is because the generated init will have `internal` access.

## Changes
- Added `package` access modifier support to `DependencyClientMacro` 
- Added unit tests verifying the change


mbrandonw (approved):
Thanks for the contribution, and thanks for adding a test!

PR #212, opened 2024-05-06
DependencyClient - Ignore @Dependency and @DependencyEndpointIgnored properties

freak4pc:
As discussed [on Slack](https://pointfreecommunity.slack.com/archives/C04L2D0MNJH/p1715002662552749), this PR adds two fixes:

1. A new `@DependencyEndpointIgnored` macro which can be used to skip a specific property from code generation
2. Automatically skip properties marked with `@Dependency` or `@DependencyEndpointIgnored`

I've added some tests and also tested it on my local use case which works well:

image


freak4pc:
FYI @stephencelis

stephencelis (approved):
Looks great! Thanks for tackling this :smile:

PR #219, opened 2024-05-29
Deadlock workaround

mbrandonw:
Right now we do a `DispatchQueue.main.sync` in order to make sure that the test observer is registered on the main thread. `sync` is always a dangerous operation, and it turns out we can deadlock. It can be reproduced with the following 3 steps:

1. Start background thread/queue and access a dependency after a small amount of time.
2. Before the dependency is accessed block the main thread with some work.
3. After the main thread work is done access another dependency.

This causes a deadlock because the background thread has already entered into a `dispatch_once` to resolve `static var current` but is currently waiting on the main thread to finish its work. And simultaneously the main thread has now also entered the `dispatch_once`, and hence a deadlock.

The fix is perhaps a little risky, but we think it should be ok. We are going to allow adding the test observer in `main.async` if the dependency is first being accessed on a background thread. It seems theoretically possible that the test observer will not be registered by the time the test ends, but we haven't been able to find a situation in which that occurs.

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -13,11 +13,11 @@ import XCTestDynamicOverlay
 
 #if _runtime(_ObjC)
   extension DispatchQueue {
-    fileprivate static func mainSync<R>(execute block: @Sendable () -> R) -> R {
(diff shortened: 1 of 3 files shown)


mbrandonw:
Incidentally this will fix #76 too.

stephencelis (approved):
Looks good!

PR #224, opened 2024-06-12
Fix some Swift 6 warnings

stephencelis:
We have a couple warnings when building in Xcode 16.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,11 +16,11 @@ concurrency:
 jobs:
   macos:
     name: macOS
-    runs-on: macos-13
+    runs-on: macos-14
     strategy:
       matrix:
         config: ['debug', 'release']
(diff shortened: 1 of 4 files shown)


stephencelis:
@finestructure While updating things for the beta, we realized that `--static-swift-stdlib` has been broken on Swift 5.9+ for awhile now. The failures are a little cryptic and we're not sure how to fix, so I'm removing it from CI for now, but if you still have a need to support it, we'd be happy to take a new PR to fix!

finestructure:
Thanks for the ping, @stephencelis ! Broken how? Because we're still using the flag in our linux builds (although in a project where we don't yet use swift-dependencies, but will very soon).

stephencelis:
@finestructure I haven't dug into it too deeply, but I think when Windows dynamic linking support was added it stopped working. If you are able to get things working, though, please let us know!

finestructure:
Will do!

PR #232, opened 2024-06-27
add throwing support to WithRandomGenerator

samisuteria:
I was using `WithRandomGenerator` with a function that that can throw and thought this might be a welcome addition for others.


stephencelis (approved):
Thanks!

PR #254, opened 2024-08-01
Add XCTestDynamicOverlay dependency to DependenciesMacros target

andriyslyusar:
In order to avoid `No such module 'XCTestDynamicOverlay'` compiler error from "DependenciesMacros/Internal/Exports.swift".


mbrandonw (approved):
Good catch, thanks @andriyslyusar !

Issue #253, opened 2024-07-31
Swift nightly snapshot compile error


Andrewangeta:
### Description

Using the swift nightly snapshots on ubuntu results in a new compile error

```
#23 19.77 /build/.build/checkouts/swift-dependencies/Sources/Dependencies/WithDependencies.swift:491:17: error: non-sendable type 'R' returned by implicitly asynchronous call to nonisolated function cannot cross actor boundary
#23 19.77 481 | #if swift(>=6)
#23 19.77 482 |   @_transparent
#23 19.77 483 |   private func isSetting<R>(
#23 19.77     |                          `- note: consider making generic parameter 'R' conform to the 'Sendable' protocol
#23 19.77 484 |     _ value: Bool,
#23 19.77 485 |     isolation: isolated (any Actor)? = #isolation,
#23 19.77     :
#23 19.77 489 |       try await DependencyValues.$isSetting.withValue(value, operation: operation)
#23 19.77 490 |     #else
#23 19.77 491 |       try await operation()
#23 19.77     |                 `- error: non-sendable type 'R' returned by implicitly asynchronous call to nonisolated function cannot cross actor boundary
#23 19.77 492 |     #endif
#23 19.77 493 |   }
```

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-dependencies/issues) or [discussion](https://github.com/pointfreeco/swift-dependencies/discussions).

### Expected behavior

Successful compilation using swift 6

### Dependencies version information

1.3.6

### Destination operating system

Ubuntu 22.04 AKA Jammy

PR #256, opened 2024-08-04
Swift 6 boundary issue fix

### LINKED ISSUE #255: Archive build failure when using Swift 6
### Description

When creating an archive build that includes swift-dependencies as an SPM package and Swift 6 specified as the build version, the archive build will fail on line 491 of WithDependencies.swift and the following error:

```
WithDependencies.swift:491:17 Non-sendable type 'R' returned by implicitly asynchronous call to nonisolated function cannot cross actor boundary
```

Creating a debug build or running unit tests is unaffected by the issue.

### Checklist

### Expected behavior

An archive build should be successfully built

### Actual behavior

The build will fail on line 491 of WithDependencies.swift

### Steps to reproduce

1. Add swift-dependencies as an SPM package
2. Set swift version to 6.0 in build settings of the project
3. Create an archive build of the project and observe the failure

1.3.6

iOS 18, vision OS 2, macOS 15

### Xcode version information

Xcode 16 beta 4

```shell
swift-driver version: 1.112.3 Apple Swift version 6.0 (swiftlang-6.0.0.6.8 clang-1600.0.23.1)
Target: arm64-apple-macosx15.0
```

hallee:
There's currently an error in release builds related to isolation of `operation`:

```481 | #if swift(>=6)
482 |   @_transparent
483 |   private func isSetting<R>(
    |                          `- note: consider making generic parameter 'R' conform to the 'Sendable' protocol
484 |     _ value: Bool,
485 |     isolation: isolated (any Actor)? = #isolation,
    :
489 |       try await DependencyValues.$isSetting.withValue(value, operation: operation)
490 |     #else
491 |       try await operation()
    |                 `- error: non-sendable type 'R' returned by implicitly asynchronous call to nonisolated function cannot cross actor boundary
492 |     #endif
493 |   }
```

I believe the correct fix for this is marking `operation` with the `sending` keyword. Here's the Swift proposal for that keyword, maybe someone who better understands concurrency should confirm that my understanding is correct: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0430-transferring-parameters-and-results.md

Fixes #255

### DIFF
--- a/Sources/Dependencies/WithDependencies.swift
+++ b/Sources/Dependencies/WithDependencies.swift
@@ -483,7 +483,7 @@ private func isSetting<R>(
   private func isSetting<R>(
     _ value: Bool,
     isolation: isolated (any Actor)? = #isolation,
-    operation: () async throws -> R
+    operation: () async throws -> sending R
   ) async rethrows -> R {
     #if DEBUG
       try await DependencyValues.$isSetting.withValue(value, operation: operation)


stephencelis:
I think the only question here is around Swift 6 snapshots being a rapidly moving target at the moment. While we're definitely open to merging this if it fixes the problem, I think we'd also like the Swift team to acknowledge if this is an expected requirement of if its boundary-checking system should infer `sending` here. Do either of you want to ask on the forums or open a Swift issue? Or would you like us to do so?

stephencelis:
@cuteiosdev Thanks! Please keep us posted! If you reply with the forum thread here I'll be sure to follow it

Andrewangeta:
This would also address #253

PR #262, opened 2024-08-20
Fix Xcode 16 beta 6 build failure

stephencelis:
Another beta, another change to boundary checking.


mbrandonw (approved):

mbrandonw:
It's still not clear what the final version of these APIs are going to look like, so we are going to merge this to fix beta 6, but we are also going to turn off Swift 6 mode until the dust settles a bit more.

mbrandonw (approved):

PR #269, opened 2024-09-04
Bucket dependency cache by Swift Testing test

stephencelis:
Currently, the global cache can bleed between test cases. This updates the underlying cache key to take the current test into account.

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -357,17 +357,13 @@ private let defaultContext: DependencyContext = {
 @_spi(Internals)
 public final class CachedValues: @unchecked Sendable {
   public struct CacheKey: Hashable, Sendable {
-    let id: ObjectIdentifier
+    let id: TypeIdentifier
     let context: DependencyContext
-#if DEBUG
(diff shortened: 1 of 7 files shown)


Sources/Dependencies/DependencyValues.swift
```diff
@@ -461,3 +477,25 @@ public final class CachedValues: @unchecked Sendable {
...
+  #if DEBUG
+  let typeName: String
+  #endif
+
+  init<T>(_ type: T.Type) {
+    self.id = ObjectIdentifier(type)
+    #if DEBUG
+    self.typeName = Dependencies.typeName(type)
+#endif
+  }
```
stephencelis, line 492: @mbrandonw Alternately, if we don't want to eagerly do formatting:

```suggestion
struct TypeIdentifier: Hashable {
  let id: ObjectIdentifier
  #if DEBUG
    let base: Any.Type
  #endif

  init<T>(_ type: T.Type) {
    self.id = ObjectIdentifier(type)
    #if DEBUG
      self.base = type
    #endif
  }
```

mbrandonw (approved):

PR #276, opened 2024-09-15
Remove experimental feature because upcoming feature 'StrictConcurrency' is already enabled as of Swift version 6

shimastripe:
## WHY

- `Package@swift-6.0.swift` enables `swiftLanguageVersions: [.v6]`
- Swift6 mode enables StrictConcurrency, so this option is not necessary.
    - Sometimes this option triggers error (ref: https://github.com/shimastripe/InAppPurchaseViewer/actions/runs/10870715912/job/30163657735 )

## WHAT

- remove `.enableExperimentalFeature("StrictConcurrency")`


stephencelis (approved):
Thanks!

PR #277, opened 2024-09-16
Update Xcode to 16 RC.

mbrandonw:
Noticed that CI started failing because there is no more beta 6 on GitHub.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         config: ['debug', 'release']
-        xcode: ['15.4', '16_beta_6']
+        xcode: ['15.4', '16.0']
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -31,8 +31,6 @@ jobs:
         run: make test-swift
       - name: Build platforms ${{ matrix.config }}
         run: CONFIG=${{ matrix.config }} make build-all-platforms
-      # - name: Build for library evolution
-      #   run: make build-for-library-evolution
```
mbrandonw, line 35: if we wanna support this we need to go all the way back to Issue Reporting to fix some things. But also I'm not sure if we really need to support this unless someone is depending on it (which clearly no one is since it's been broken for awhile).

stephencelis (approved):

PR #279, opened 2024-09-24
Restore Library Evolution support for Xcode 16

gemmakbarlow:
- Builds for Library Evolution support, and patches a small issue in this repository causing an issue for us, i.e. - 

image

- Adds Library Evolution support target replicating the issue to CI

**Additional Notes**
I noticed [this comment](https://github.com/pointfreeco/swift-dependencies/pull/277#discussion_r1761462481); we're needing Library Evolution support for some beta apps we're working on using `swift-dependencies`. 

Happy to assist with the maintenance, as far as it useful !

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ concurrency:
 jobs:
   macos:
     name: macOS
-    runs-on: macos-latest
+    runs-on: macos-14
     strategy:
       matrix:
         config: ['debug', 'release']
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ concurrency:
 jobs:
   macos:
     name: macOS
-    runs-on: macos-14
+    runs-on: macos-latest
     strategy:
       matrix:
         config: ['debug', 'release']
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ concurrency:
 jobs:
   macos:
     name: macOS
-    runs-on: macos-14
+    runs-on: macos-latest
```
stephencelis, line 19: I think this resolves to `macos-15`, which may be fine to explore using, but we've found that the newer runners are usually less available when first introduced. Is it possible to stick with `macos-14` for now or is `macos-15` required for this PR? (If required, can we update this to be explicit so that CI doesn't automatically upgrade in the future?)

stephencelis (commented):
Thanks for continuing to take the time to PR these!

.github/workflows/ci.yml
gemmakbarlow: Totally fine, `macos-14` it shall remain.

Makefile
```diff
@@ -50,6 +50,13 @@ build-for-library-evolution:
 		-Xswiftc -emit-module-interface \
 		-Xswiftc -enable-library-evolution
 
+	swift build \
```
gemmakbarlow, line 53: Noting that this really only covers Library Evolution on `macOS`, but is an improvement on what's currently setup for CI.

stephencelis:
Thanks!

PR #293, opened 2024-10-16
Override Test ID to ensure trait dependencies are cached

jflan-dd:
I ran into an issue where reference typed dependencies weren't correctly cached if they were updated in-place rather than being fully overridden.

The root cause of this seems to be that `Test.current` isn't set during `prepare(for test: Test)` and as a result the `CacheKey.testIdentifier` was `nil` during dependency setup.

~~This PR creates an internal `TestID` struct that can be explicitly overridden during dependency prep, but perhaps there's a more elegant solution that could be implemented in `xctest-dymaic-overlay` directly.~~

This PR is dependent on https://github.com/pointfreeco/swift-issue-reporting/pull/137

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -362,31 +362,6 @@ private let defaultContext: DependencyContext = {
 }()
 
 @_spi(Internals)
-public struct TestID: @unchecked Sendable {
-  @TaskLocal private static var override: Self?
-  
-  public let value: AnyHashable
-  
-  public init(_ value: AnyHashable) {
-    self.value = value
-  }
-  
-  public static func withCurrent<R>(_ testID: AnyHashable, perform body: () async throws -> R) async rethrows -> R {
-    try await Self.$override.withValue(TestID(testID), operation: body)
-  }
-  
-  public static var current: Self? {
-    if let value = override { return value }
-    
-    if case let .swiftTesting(.some(testing)) = TestContext.current {
-      return .init(testing.test.id.rawValue)
-    }
-    
-    return nil
-  }
-}
-
-@_spi(Internals)
 public final class CachedValues: @unchecked Sendable {
   @TaskLocal static var isAccessingCachedDependencies = false
 
@@ -398,10 +373,10 @@ public final class CachedValues: @unchecked Sendable {
     init(id: TypeIdentifier, context: DependencyContext) {
       self.id = id
       self.context = context
-      if let testID = TestID.current {
-        // unsafeBitCast used because current version of xctest-dynamic-overlay does not expose init
-        self.testIdentifier = unsafeBitCast(testID, to: TestContext.Testing.Test.ID.self)
-      } else {
+      switch TestContext.current {
+      case let .swiftTesting(.some(testing)):
+        self.testIdentifier = testing.test.id
+      default:
         self.testIdentifier = nil
       }
     }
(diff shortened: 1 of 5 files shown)


jflan-dd:
It seems like this could also be solved with a `TestExecuting` conformance if https://forums.swift.org/t/pitch-custom-test-execution-traits/75055/14 lands, since `Test.current` is set before `_executeTraits` is called:
https://github.com/swiftlang/swift-testing/blob/751ea94cc79b0f27c67ac88db4d348ad8703ae1a/Sources/Testing/Running/Runner.swift#L201-L203

Is there a plan for how to be backwards compatible with different versions of `swift-testing` as new features are added?

Sources/Dependencies/DependencyValues.swift
```diff
@@ -373,10 +398,10 @@ public final class CachedValues: @unchecked Sendable {
     init(id: TypeIdentifier, context: DependencyContext) {
       self.id = id
       self.context = context
-      switch TestContext.current {
-      case let .swiftTesting(.some(testing)):
-        self.testIdentifier = testing.test.id
-      default:
+      if let testID = TestID.current {
+        // unsafeBitCast used because current version of xctest-dynamic-overlay does not expose init
+        self.testIdentifier = unsafeBitCast(testID, to: TestContext.Testing.Test.ID.self)
```
jflan-dd, line 403: This hack is the main reason I marked this PR as a draft

stephencelis:
@jflan-dd Thanks for looking into this and putting together a PR! We'll try to review it soon.

jflan-dd:
@stephencelis I created a [separate PR](https://github.com/pointfreeco/swift-issue-reporting/pull/137) in `swift-issue-reporting` to support overriding the TestContext which greatly simplified this PR

stephencelis:
@jflan-dd This hasn't fallen off our radar but we haven't had time to fully review and grok things yet. Hopefully soon!

jflan-dd:
@stephencelis no problem. We're working around this issue by explicitly setting the mock dependency inside of the update block so that the `DependencyValues` setter is called directly, although I haven't tested our workaround with [this change](https://github.com/pointfreeco/swift-dependencies/pull/287) yet.

For additional context my understanding of the issue is that because `swift-testing` doesn't set the test ID before `prepare(for:)` it's causing [this cache key](https://github.com/pointfreeco/swift-dependencies/blob/a24a0f1037b5c65885876be89f598b6f1e70f21c/Sources/Dependencies/DependencyValues.swift#L478) to have a `nil` `testIdentifier` when updating dependencies inside of the `updateValues` closure. This isn't an issue for dependencies with value semantic because they end up in [this else case](https://github.com/pointfreeco/swift-dependencies/blob/a24a0f1037b5c65885876be89f598b6f1e70f21c/Sources/Dependencies/DependencyValues.swift#L336-L338) which stores the updated dependency directly in the `DependencyValues` `storage`.

This results in caching the updated dependency with an invalid cache key, which causes a cache miss when the dependency is accessed during the test.

This image shows a breakpoint while accessing the dependency within the body of a test. The cache contains a dependency with the same `TypeIdentifier` from the trait update, but there's a cache miss because the `testIdentifiers` don't match.
image

stephencelis:
@jflan-dd Any reason for closing this? It hasn't been on the top of our priority list but we would like to track any issues with the library that should be fixed.

jflan-dd:
@stephencelis I haven't actually tested it, but my assumption was that test scoping would address this

jflan-dd:
@stephencelis I just confirmed that the tests from this PR pass on `main` with Swift 6.1 using test scoping, but fail with Swift 6.0.

I'm happy to re-open the PR if you'd like to consider it for the Swift 6.0 codepath.

Issue #302, opened 2024-11-12
1.5.0 fails to build on Linux


finestructure:
### Description

```
root@orbstack:/host# git rev-parse @
a24a0f1037b5c65885876be89f598b6f1e70f21c
root@orbstack:/host# swift --version
Swift version 6.0.2 (swift-6.0.2-RELEASE)
Target: aarch64-unknown-linux-gnu
root@orbstack:/host# swift build
warning: 'combine-schedulers': /host/.build/checkouts/combine-schedulers/Package@swift-6.0.swift:5:15: warning: 'init(name:defaultLocalization:platforms:pkgConfig:providers:products:dependencies:targets:swiftLanguageVersions:cLanguageStandard:cxxLanguageStandard:)' is deprecated: replaced by 'init(name:defaultLocalization:platforms:pkgConfig:providers:products:dependencies:targets:swiftLanguageModes:cLanguageStandard:cxxLanguageStandard:)'
 3 | import PackageDescription
 4 | 
 5 | let package = Package(
   |               |- warning: 'init(name:defaultLocalization:platforms:pkgConfig:providers:products:dependencies:targets:swiftLanguageVersions:cLanguageStandard:cxxLanguageStandard:)' is deprecated: replaced by 'init(name:defaultLocalization:platforms:pkgConfig:providers:products:dependencies:targets:swiftLanguageModes:cLanguageStandard:cxxLanguageStandard:)'
   |               `- note: use 'init(name:defaultLocalization:platforms:pkgConfig:providers:products:dependencies:targets:swiftLanguageModes:cLanguageStandard:cxxLanguageStandard:)' instead
 6 |   name: "combine-schedulers",
 7 |   platforms: [
warning: 'swift-clocks': /host/.build/checkouts/swift-clocks/Package@swift-6.0.swift:5:15: warning: 'init(name:defaultLocalization:platforms:pkgConfig:providers:products:dependencies:targets:swiftLanguageVersions:cLanguageStandard:cxxLanguageStandard:)' is deprecated: replaced by 'init(name:defaultLocalization:platforms:pkgConfig:providers:products:dependencies:targets:swiftLanguageModes:cLanguageStandard:cxxLanguageStandard:)'
 3 | import PackageDescription
 4 | 
 5 | let package = Package(
[...]

PR #303, opened 2024-11-12
Add Swift 6.0 to Linux test matrix

finestructure:
This should show the issue in #302


stephencelis (approved):
Thanks!

mbrandonw:
Hi @finestructure, following up on this, looking at the CI logs it seems that Ubuntu compiled just fine. The only thing that is failing are some tests that use the `.dependency` test trait. I'm not sure why the test traits don't work on Linux, but at least the library does seem to compile.

Are you still having problems with this?

finestructure:
Hi @mbrandonw ! As I recall we fixed the immediate problem, the build failure, via https://github.com/pointfreeco/swift-dependencies/pull/304. I believe main point of this PR was, via https://github.com/pointfreeco/swift-dependencies/pull/303/commits/6535dabe92d64a972f44f2a30ce11d23ec834a07, to have Linux/Swift 6 in your test matrix because that's our prod environment :)

PR #305, opened 2024-11-12
Fix OpenURL crash on macOS Sequoia

beetee17:
Refer to the discussion at #301 for details.

### DIFF
--- a/Sources/Dependencies/DependencyValues/OpenURL.swift
+++ b/Sources/Dependencies/DependencyValues/OpenURL.swift
@@ -12,7 +12,7 @@
 
   @available(iOS 14, macOS 11, tvOS 14, watchOS 7, *)
   private enum OpenURLKey: DependencyKey {
-    static let liveValue = OpenURLEffect { url in
+    static let liveValue = OpenURLEffect { @MainActor url in
       let stream = AsyncStream<Bool> { continuation in
         let task = Task { @MainActor in
           #if os(watchOS)


Sources/Dependencies/DependencyValues/OpenURL.swift
```diff
@@ -12,7 +12,7 @@
 
   @available(iOS 14, macOS 11, tvOS 14, watchOS 7, *)
   private enum OpenURLKey: DependencyKey {
-    static let liveValue = OpenURLEffect { url in
+    static let liveValue = OpenURLEffect { @MainActor url in
```
stephencelis, line 15:
```suggestion
    // NB: This must be annotated '@MainActor' to avoid a Swift bug:
    //     https://github.com/pointfreeco/swift-dependencies/discussions/301
    static let liveValue = OpenURLEffect { @MainActor url in
```

stephencelis:
@beetee17 Thanks, and sorry for the delay! This time of year has been busy.

Do you think it's possible to get a unit test in place that demonstrates the problem? Ideally we would have CI catch any regressions if someone were to refactor away that `@MainActor` annotation. I've at the very least added a suggestion to comment on the inclusion of it.

beetee17:
@stephencelis, sorry for the delay as well. I was finally able to gain access to macOS Sequoia 15.0 on an M1 Mac Mini for testing. I was able to verify the root cause of the crash, however, the fix comes with a drawback. 

The following is copy-pasted from the [latest commit message](https://github.com/pointfreeco/swift-dependencies/pull/305/commits/7e6f8098940ffeb4b60e354fed82fe74d2af9b90), let me know if more details are required:

> Further testing has revealed that the crash stems from the completion handler for `openURL`. This may be a bummer for use-cases that require it to check whether an app was available to handle a URL scheme.
>
> An attempt was made to have the best of both worlds by returning `.systemAction` in the `OpenURLAction` handler, however this causes the completion handler to never return.
>
> Therefore, the only real fix is to always return `.handled`, which makes it equivalent to the watchOS implementation that always yields `true`. 

The unit test is simple, it simply tests the live value and ensures that the function returns without crashing. However, since the live implementation was hidden behind the private enum `OpenURLKey`, I changed the access modifier to `internal` for the purpose of testing.

mansbernhardt:
I'm experiencing this crash as well. @stephencelis what is the status on this fix?

PR #309, opened 2024-11-15
Fix state sharing bug in `.dependency(_:_:)` trait

TimHendricksenDD:
When the `.dependency(_:_:)` function is used with a reference type to override the dependency for a test suite, tests can inadvertently share state because the same instance is used for each test:
```swift
@Suite(.serialized, .dependency(\.classClient, ClassClient()))
struct Example {
  @Test
  func test1() {
    @Dependency(\.classClient) var client
    client.count += 1
    #expect(client.count == 1)
  }

  @Test
  func test2() {
    @Dependency(\.classClient) var client
    client.count += 1
    #expect(client.count == 1) // 🛑 Expectation failed: (client.count → 2) == 1
  }
}
```

To fix this, this PR changes the `value` argument to be an `@autoclosure` so that a new instance can be created for each test. Let me know if there's an alternate solution that would be better!

### DIFF
--- a/Sources/DependenciesTestSupport/TestTrait.swift
+++ b/Sources/DependenciesTestSupport/TestTrait.swift
@@ -43,10 +43,10 @@
     ///   - value: A dependency value to override for the test.
     public static func dependency<Value>(
       _ keyPath: WritableKeyPath<DependencyValues, Value> & Sendable,
-      _ value: sending Value
+      _ value: @escaping @Sendable @autoclosure () -> Value
     ) -> Self {
-      Self { [uncheckedValue = UncheckedSendable(value)] in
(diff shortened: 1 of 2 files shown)


stephencelis (approved):
Good catch, thanks!

TimHendricksenDD:
Fixed in https://github.com/pointfreeco/swift-dependencies/pull/314

stephencelis:
@TimHendricksenDD Sorry that we lost track of this fix and ended up reimplementing! Brandon and I have been traveling and it's been harder to keep track of everything in flight. Thanks for taking the time to dig into the problem and submit a patch all the same.

TimHendricksenDD:
That's ok - thanks for following up!

PR #315, opened 2024-12-03
Fix prepareDependencies when using TestDependencyKey.

mbrandonw:
Currently `prepareDependencies` does not work for dependencies that only have a `TestDependencyKey` conformance. This fixes that.

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -480,7 +480,7 @@ public final class CachedValues: @unchecked Sendable {
         if
           context == .live,
           !DependencyValues.isSetting,
-          (cached[cacheKey] == nil || cached[cacheKey]?.preparationID == nil),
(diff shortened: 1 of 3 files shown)


Sources/Dependencies/DependencyValues.swift
```diff
@@ -477,7 +477,12 @@ public final class CachedValues: @unchecked Sendable {
     return withIssueContext(fileID: fileID, filePath: filePath, line: line, column: column) {
       let cacheKey = CacheKey(id: TypeIdentifier(key), context: context)
       #if DEBUG
-        if context == .live, !DependencyValues.isSetting, !(key is any DependencyKey.Type) {
+        if
+          context == .live,
+          !DependencyValues.isSetting,
+          !(cached[cacheKey] != nil && cached[cacheKey]?.preparationID != nil),
```
mbrandonw, line 483: We do not need to warn if there is something in the cache that came from preparing dependencies.

stephencelis (approved):

PR #319, opened 2024-12-18
Update README.md

vladpetruk:


stephencelis:
This actually is the intended tense, but thanks for taking the time to PR :smile:

If it's a cause of confusion again in the future we can consider rewriting the sentence differently.

PR #321, opened 2024-12-23
Don't cache preview app entry point dependencies

stephencelis:
We have a longstanding gotcha where we tell folks to avoid instantiating their root model on their app entry point because any dependencies on the root model may negatively affect any preview in their application.

This PR detects when a dependency is accessed in a SwiftUI preview app entry point and prevents it from influencing the cache using the call stack symbols available. While this isn't a silver bullet, and any async work kicked off from the app entry point could still affect things negatively, this should hopefully be mostly an improvement on the status quo and maybe we won't have to refer folks to this gotcha as often in the future.

### DIFF
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -269,7 +269,7 @@ public struct DependencyValues: Sendable {
     set {
       if DependencyValues.isPreparing {
         if context == .preview, Thread.isPreviewAppEntryPoint {
-          reportIssue("Ignoring dependencies prepared in app entry point")
+          reportIssue("Ignoring dependencies prepared in preview app entry point")
           return
         }
         let cacheKey = CachedValues.CacheKey(id: TypeIdentifier(key), context: context)
--- a/Sources/Dependencies/DependencyValues.swift
+++ b/Sources/Dependencies/DependencyValues.swift
@@ -268,6 +268,10 @@ public struct DependencyValues: Sendable {
     }
     set {
       if DependencyValues.isPreparing {
+        if context == .preview, Thread.isPreviewAppEntryPoint {
+          reportIssue("Ignoring dependencies prepared in app entry point")
+          return
+        }
         let cacheKey = CachedValues.CacheKey(id: TypeIdentifier(key), context: context)
         guard !cachedValues.cached.keys.contains(cacheKey) else {
           if cachedValues.cached[cacheKey]?.preparationID != DependencyValues.preparationID {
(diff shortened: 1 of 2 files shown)


mbrandonw (approved):

stephencelis:
- [ ] Update Gotchas doc

niorko:
@stephencelis @mbrandonw 

We’ve encountered an issue with stateful `testValue` in tests when running test suites in parallel. It seems that `testValue` is being cached, which causes shared state between tests. To address this, we wrapped each test case with `withDependencies` and overrode the dependency explicitly. However, this adds unnecessary boilerplate.

This behavior appears to be related to caching introduced in PR #78, though I may be mistaken. Ideally, testValue should not be cached and should reset automatically for each test case, ensuring isolation even during parallel execution.

Could we consider revisiting this behavior to ensure testValue is stateless across tests? This would align better with typical testing practices. Thank you.

MasekHJ:
Adding to @niorko comment:

To be more specific, it seems that issue is that Swift-Testing no longer resets cache.

XCTest previously executed this block

```
        let testCaseWillStartBlock: @convention(block) (AnyObject) -> Void = { _ in
          DependencyValues._current.cachedValues.cached = [:]
        }
```
        
This no longer applies to Swift-Testing.
But even with cache reset in our findings it would not work with parallel run:

```
final class MultipleParallelTests {
    
    init() {
        DependencyValues._current.resetCache()
    }
}
```

Because some `Dependencies` in the code were resolved before cache was reset (maybe like race condition) due to being run in parallel.

using this will work - but this is just a workaround.

```
@Suite(.serialized)
final class MultipleSerialTests {
    
    init() {
       DependencyValues._current.resetCache()
    }
}
  ````
  
  Is there a possibility that Swift Testing @Test methods could be isolated from each other (use own DependencyValues / cache) when being run in parallel?
  
  Of course as @niorko mentioned we can add `withDependencies` to every test and keep the nice-to-have parallel testing, but this adds boilerplate and developers have to think what dependencies to override, because they will be cached (maybe unexpectedly)

niorko:
We were using an older version that lacked support for Swift Testing. After updating, it appears to work correctly - apologies for the earlier comments. However, it seems to have some issues with repeated tests, which are likely demonstrated here https://github.com/pointfreeco/swift-dependencies/pull/269/commits/7d89478e4f0392f7d36ddf68cad55b3deae67a87#diff-86f2ce713ec4eb2b2a5301f1b558597d3d362f26a1152c66fb117e2580402c2a.

PR #323, opened 2024-12-30
Deprecate preview dependencies trait

stephencelis:
Due to a quirk or bug or Xcode previews, dependency traits are executed for every preview in a file, not just the one that's running. This can cause dependencies from other previews to trample over the preview you are viewing.

Because of this, let's deprecate this functionality in favor of good ole fashioned `withDependencies`, with the option of using `prepareDependencies`.


Sources/Dependencies/Dependency.swift
```diff
@@ -233,19 +243,19 @@ extension DependencyValues {
...
       self[
         Key.self,
-        key.fileID,
-        key.filePath,
-        key.line,
-        key.column
+        fileID: key.fileID,
+        filePath: key.filePath,
+        line: key.line,
+        column: key.column
```
stephencelis, line 249: Noticed the lack of labels when fixing docs. This is technically breaking but I don't think anyone should be invoking these arguments directly outside the library and there's a fix-it to add the labels, but if we're worried I can undo this.

Sources/Dependencies/Internal/AppEntryPoint.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 extension Thread {
-  public static var isPreviewAppEntryPoint: Bool {
```
stephencelis, line 4: Accidentally made this public in the previous release.

mbrandonw (approved):

PR #330, opened 2025-01-14
Fix build of unit tests on Xcode 16

pyrtsa:
The unit test file `TestTraitTests.swift` failed to build on Xcode 16:

> ```
> 'sleep(for:tolerance:clock:)' is only available in iOS 16.0 or newer
>
> 'milliseconds' is only available in iOS 16.0 or newer
> ```

Adding the appropriate `@available` markers fixes the build.


mbrandonw (approved):
Hi @pyrtsa, good catch! Thanks for that. Looks like we only run tests on macOS currently on CI.

PR #340, opened 2025-02-21
Enable creating initializers even without endpoints

fruitcoder:
**Motivation**
In our projects we like to group endpoint dependencies in their own @DependencyClient domains and then expose them via a proxy struct that just has the sub dependencies as members. We'd like to just add the @DependencyClient macro to the proxy struct and get the initializer creation for free. 

**Impact**
It's a subtle change to the end of the macro which doesn't early out when there's no endpoint members

**Detailed example**
We usually have to following setup with api clients or database clients:
```swift
// Package APIClient
@DependencyClient
public struct APIClientBookmarks {
  public var fetchBookmarks: @Sendable () async -> [Bookmark] = { [] } 
  // many more endpoints
}
      
@DependencyClient
public struct APIClientVideos {
  public var fetchVideos: @Sendable () async -> [Video] = { [] } 
}
     
@DependencyClient
public struct APIClient {
  public var bookmarks: APIClientBookmarks
  public var videos: APIClientVideos
}
```

Resulting macro expansion from the PR (domain specific dependency client macro expansion is unchanged)
```swift
public struct APIClient {
  public var bookmarks: APIClientBookmarks
  public var videos: APIClientVideos
      
  public init(
    bookmarks: APIClientBookmarks,
    videos: APIClientVideos
  ) {
    self.bookmarks = bookmarks
    self.videos = videos
  }
}
```

Usage in feature code
```swift
class FeatureUsingVideoEndpoint {
  @Dependency(\.apiClient.videos) var videoClient
}
```

### DIFF
--- a/Sources/DependenciesMacrosPlugin/DependencyClientMacro.swift
+++ b/Sources/DependenciesMacrosPlugin/DependencyClientMacro.swift
@@ -216,10 +216,12 @@ public enum DependencyClientMacro: MemberAttributeMacro, MemberMacro {
       )
       hasEndpoints = hasEndpoints || isEndpoint
     }
-    guard hasEndpoints else { return [] }
     let access = accesses.min().flatMap { $0.token?.with(\.trailingTrivia, .space) }
+    let propertyCandidates = hasEndpoints
+      ? [properties, properties.filter { !$0.isEndpoint }]
+      : [properties.filter { !$0.isEndpoint }]
     // TODO: Don't define initializers if any single endpoint is invalid
-    return [properties, properties.filter { !$0.isEndpoint }].map {
+    return propertyCandidates.map {
       $0.isEmpty
         ? "\(access)init() {}"
         : """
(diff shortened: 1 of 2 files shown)


stephencelis:
@fruitcoder Thanks for sharing and exploring a PR. So is this equivalent to not applying the macro and relying on the memberwise synthesized initializer (or defining a public initializer yourself)?

I think our main worry is that the purpose of `@DependencyClient` is instrument closure endpoints, and so making it also a public initializer generator muddies that purpose a bit. Have you looked at https://github.com/gohanlon/swift-memberwise-init-macro as a more general purpose solution to generating this initializer?

fruitcoder:
@stephencelis you're right, it's just a tiny convenience not to have to re-type `ini..` and adding `public` when adding new domains but we thought it might be a similar annoyance for other folks who use the same structure. 
We are aware of the `swift-memberwise-init-macro` (and others) but didn't want to add an extra dependency for this tiny nitpickiness of ours 😄 
Hopefully Swift will introduce more macros to handle cases like automatically publicizing struct initializers in the future. Feel free to close this PR!

stephencelis:
@fruitcoder Thanks regardless for exploring! We'll continue to think about it!

PR #343, opened 2025-02-25
Add dependencies provider

finestructure:
When using the existing `@Trait(.dependencies(...))` there is no way to "house" a set of dependencies on a type in order to group them together and make them available via `.` + autocomplete, like for dependency keys.

In order to group them, I'm resorting to a top level private function:

```swift
private let defaultDependencies: @Sendable (inout DependencyValues) -> Void = {
    $0.logger.log = { @Sendable _, _ in }
    $0.shell = .liveValue
}
```

which then at least allows

```swift
    @Test(.dependencies(defaultDependencies))
    func commitCount() async throws {
        try await withGitRepository { path in
            #expect(try await Git.commitCount(at: path) == 57)
        }
    }
```

With the proposed change, I could write this as

```swift
  extension DependenciesProvider {
    static var `default`: Self {
      .init {
         $0.logger.log = { @Sendable _, _ in }
         $0.shell = .liveValue
      }
    }
  }
```

and

```swift
    @Test(.dependencies(.default))
    func commitCount() async throws {
        try await withGitRepository { path in
            #expect(try await Git.commitCount(at: path) == 57)
        }
    }
```

It might not look like a huge difference, but being able to house the dependencies in the `DependenciesProvider` type would allow for succinct use and re-use of them across the test base.

The alternatives all feel a bit clunky:

- top level functions litter the scope and aren't very discoverable
- creating a new type with statics would avoid the top level litter but require referencing of the type each time, i.e. `@Trait(.dependencies(TestDependencies.default))`

Would this be a worthwhile addition? Or is there another way to achieve what I'm trying to do that I'm not seeing?

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "46c7c52f0c1617cc1d5bc47663541c21a6e6734ddf2ad859bf5bf5bc207fa8f4",
+  "originHash" : "9af75795d44ca33d2d2ff1228ae0b2082fad28fb82fe8d750bbc2a258029bc84",
   "pins" : [
     {
       "identity" : "combine-schedulers",
@@ -78,8 +78,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/pointfreeco/xctest-dynamic-overlay",
       "state" : {
-        "revision" : "770f990d3e4eececb57ac04a6076e22f8c97daeb",
-        "version" : "1.4.2"
+        "branch" : "fix-linux-windows",
+        "revision" : "64a5e4f4bf763fa9a9ed7757ae656fa11aadccac"
       }
     }
   ],
--- a/Package@swift-6.0.swift
+++ b/Package@swift-6.0.swift
@@ -29,7 +29,7 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/combine-schedulers", from: "1.0.2"),
     .package(url: "https://github.com/pointfreeco/swift-clocks", from: "1.0.4"),
     .package(url: "https://github.com/pointfreeco/swift-concurrency-extras", from: "1.0.0"),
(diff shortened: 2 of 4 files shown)


stephencelis:
An alternative would be to extend the trait directly:

```swift
extension Trait where Self == _DependenciesScopeTrait {
  static var defaultDependencies: Self {
    .dependencies {
      $0.logger.log = { @Sendable _, _ in }
      $0.shell = .liveValue
    }
  }
}

@Suite(.defaultDependencies) struct BaseSuite {}
```

But also I think you're showing that this could be defined locally in your project or even as a new package that extends this one, so that's nice at the very least.

We'll chat and think about it!

finestructure:
I think you meant `_DependenciesTrait`, right? That would work, although I find the scoping with `.dependencies(.default)` a bit nicer.

The reason I opened the PR is because I can't currently do what I propose locally, because the `_DependenciesTrait` isn't public

```
    package init(_ updateValues: @escaping @Sendable (inout DependencyValues) throws -> Void) {
      self.updateValues = updateValues
    }
```

That led me to see if it works by making the change upstream and then I thought I might as well PR it :)

So guess a more limited version of this would be just to make that initialiser `public`? Is that something you'd consider?

stephencelis:
> The reason I opened the PR is because I can't currently do what I propose locally, because the `_DependenciesTrait` initialiser isn't public

Could you use the static `dependencies` function, instead?

finestructure:
No worries if this isn't something you want to add at this point. Feel free to close the PR, I don't want to clutter up your repo :)

stephencelis:
@finestructure Sorry, I mean can you do the following outside the library?

```swift
 extension Trait where Self == _DependenciesTrait {
   public static func dependencies(
     _ provider: DependenciesProvider
   ) -> Self {
-    Self(provider.updateValues)
+    .dependencies(provider.updateValues)
   }
 }
```

finestructure:
Ah, of course! That works, thanks for the suggestion.

I'll close this then since I can do all this on my end now :)

PR #345, opened 2025-02-25
responsibly -> responsible in Testing article

agarrharr:
I was reading the new documentation and I thought this word should be "responsible" instead of "responsibly".


stephencelis (approved):
Thanks!

Issue #349, opened 2025-03-11
`@DependencyEndpoint` is not using ACL keyword for the method equivalent when using `package` keyword


gsl-sebastienpecoul:
### Description

Hi there, 
When using the `@DependencyClient` macro on struct properties with `package` ACL the generated method with `@DependencyEndpoint` will omit the `package` keyword making it not available from other products of the package.

The issue was reported [here](https://pointfreecommunity.slack.com/archives/C04L2D0MNJH/p1741004147856899)

I've submitted a fix PR [there](https://github.com/pointfreeco/swift-dependencies/pull/350)
(Thanks @gibachan for pin pointing the issue 🙏 )
 

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [x] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-dependencies/issues) or [discussion](https://github.com/pointfreeco/swift-dependencies/discussions).

### Expected behavior
```swift
package struct Check {
    @DependencyEndpoint
    package var fetchValue: (_ id: Int) throws -> Int
}

// Should expand too
...
package func fetchValue(id p0: Int) throws -> Int { 
    try self.fetchValue(p0)
}
```

### Actual behavior
```swift
package struct Check {
    @DependencyEndpoint
    package var fetchValue: (_ id: Int) throws -> Int
}

// Will expand too
...
func fetchValue(id p0: Int) throws -> Int { // ❌ No package ACL
    try self.fetchValue(p0)
}
```

### Steps to reproduce

You can use this sample [project](https://github.com/gsl-sebastienpecoul/swift-dependencies-acl-issue)  where I showcase the issue with two simple structs.

### Dependencies version information

1.8.1

### Xcode version information

Version 16.1 (16B40)

### Swift Compiler version information

```shell
swift-driver version: 1.115 Apple Swift version 6.0.2 (swiftlang-6.0.2.1.2 clang-1600.0.26.4)
Target: arm64-apple-macosx15.0
```

mbrandonw:
Hi @gsl-sebastienpecoul, thanks for the report! Would you be interested in opening a PR to fix the problem?

gsl-sebastienpecoul:
Hi @mbrandonw - Sure! Here you go: https://github.com/pointfreeco/swift-dependencies/pull/350

PR #350, opened 2025-03-11
fix: Use proper access control when generating method with `package` ACL

gsl-sebastienpecoul:
Context -> #349


mbrandonw (approved):
Thank you @gsl-sebastienpecoul! Looks good...

PR #353, opened 2025-03-19
Undeprecate preview traits?

stephencelis:
I believe I stumbled onto a fix for preview code bleeding between previews, so this works in a single file:

```swift
struct MyView: View {
  @Dependency(\.date.now) var now
  var body: some View {
    Text("\(now)")
  }
}

#Preview(
  "Controlled",
  traits: .dependency(\.date.now, Date(timeIntervalSince1970: 0)
) {
  MyView()
}

#Preview("Uncontrolled") {
  MyView()
}
```

The one caveat to this is that if you access a dependency directly in the `#Preview` block you must use `@Previewable @Dependency`, but that's the same for `@Environment`.


Sources/Dependencies/Traits/PreviewTrait.swift
```diff
@@ -1 +1,34 @@
...
+  }
+
+  private struct DependenciesPreviewModifier: PreviewModifier {
+    let updateValuesForPreview: @Sendable (inout DependencyValues) -> Void
+
+    func body(content: Content, context: ()) -> some View {
+      prepareDependencies(updateValuesForPreview)
+      return content
+    }
+  }
```
stephencelis, line 33: Introducing a preview modifier allows us to delay when dependencies are prepared to the computation of the preview's body.

mbrandonw (approved):

Issue #183, opened 2024-02-09
When using `@DependencyClient` macro, closures with mandatory return don't use the unimplemented version on tests


gpambrozio:
### Description

When using `@DependencyClient` macro in a client that has closures with mandatory returns, when using these closures in a test without overriding them we don't get errors about the method not being implemented.

According to [this message from Brandon](https://pointfreecommunity.slack.com/archives/C04L2D0MNJH/p1707325357148669?thread_ts=1707313605.968439&cid=C04L2D0MNJH) this is related to a [swift bug](https://github.com/apple/swift/issues/71070). Given that I'd suggest to at least add a note to the documentation mentioning that, instead of having a closure such as `{ _ in true }`, we unfortunately still have to use `{ _ in unimplemented(placeholder: true }`

See also [this Slack thread](https://pointfreecommunity.slack.com/archives/C04L2D0MNJH/p1705952998951619)

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/swift-dependencies/issues) or [discussion](https://github.com/pointfreeco/swift-dependencies/discussions).

### Expected behavior

The test shown should have 2 errors.

### Actual behavior

The test only produces one error.

### Steps to reproduce

Simple Client:
```
import Dependencies
import DependenciesMacros

@DependencyClient
struct SimpleClient {
    var noReturn: () -> Void
    var withBoolReturn: () -> Bool = { true }
}

extension SimpleClient: TestDependencyKey {
    public static let testValue = SimpleClient()
}

extension DependencyValues {
    var simpleClient: SimpleClient {
        get { self[SimpleClient.self] }
        set { self[SimpleClient.self] = newValue }
    }
}
```

This test should have 2 failures but only shows one:
```
import Dependencies
import XCTest
@testable import TCATest

final class TCATestTests: XCTestCase {
    func testExample() throws {
        @Dependency(\.simpleClient) var simpleClient
        _ = simpleClient.withBoolReturn()
        simpleClient.noReturn()
    }
}
```

### Dependencies version information

1.2.1

### Destination operating system

iOS

### Xcode version information

15.2

### Swift Compiler version information

```shell
[...]

PR #355, opened 2025-04-05
Resolves issue with endpoints not reporting errors when there's a custom closure

gpambrozio:
Resolves issue #183

After [getting feedback from the swift team](https://github.com/swiftlang/swift/issues/71070#issuecomment-2780647105) on an issue that seems to be on the swift side I modified the endpoint macro to follow their advice. Simply removing the member initializer accessor.

I also modified all tests so they pass. 

I tested on a project that uses dependencies a lot and it all still works. I also tested removing some `unimplemented` from my closures and checking if tests fail and they do.

### DIFF
--- a/Sources/DependenciesMacrosPlugin/DependencyEndpointMacro.swift
+++ b/Sources/DependenciesMacrosPlugin/DependencyEndpointMacro.swift
@@ -32,12 +32,6 @@ public enum DependencyEndpointMacro: AccessorMacro, PeerMacro {
 
     return [
       """
-      @storageRestrictions(initializes: _\(raw: identifier))
-      init(initialValue) {
-      _\(raw: identifier) = initialValue
-      }
-      """,
-      """
       get {
       _\(raw: identifier)
       }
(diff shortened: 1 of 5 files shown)


mbrandonw (approved):
Thanks for taking care of this so quickly @gpambrozio!

AgapovOne:
This actually breaks some dependencies that were declared before, i.e.:

```
struct Client {
  var someProperty: Int
  @DependencyEndpoint
  var endpoint: () -> Void
}

public func doX() {

  let client1 = Client(someProperty: 0, _endpoint: { print("doY") })
  let client2 = Client(someProperty: 0, endpoint: { print("doY") })
}
```

I could have used init before for client2, now both inits break:

<details>
<summary>Screenshot</summary>

Screenshot 2025-04-22 at 14 18 01

</details>

Only custom init would help here, where i should write `self._endpoint = endpoint`. I would prefer if init with `endpoint` name without underscore is used still, because that macro alone helps in some places where we want to use functions.

AgapovOne:
Here's an example of dependency where I can't use @DependencyClient, because I don't know a default for ValueType:

```
struct SettingWritable<ValueType: Codable & Equatable & Sendable> {
  public var value: () -> ValueType
  @DependencyEndpoint
  public var set: (ValueType?) -> Void
}
```

I'll try to look into macro arguments and somehow fix this. Good idea with boolean parameter, I'll try!

PR #357, opened 2025-04-09
chore: Remove no longer relevant warning in Designing dependencies documentation page

gsl-sebastienpecoul:
Hi!
I saw that in the 1.9.1 a fix was done in #355 to don't have to specify `reportIssue` anymore in non throwing closures.
This case was mentioned in the documentation [here](https://swiftpackageindex.com/pointfreeco/swift-dependencies/1.9.1/documentation/dependencies/designingdependencies#DependencyClient-macro) 

CleanShot 2025-04-09 at 16 06 33

Since it's was fixed I think we can now remove this part from the documentation.


mbrandonw (approved):
Good catch @gsl-sebastienpecoul, thanks!

PR #362, opened 2025-04-23
Add init accessor to DependencyEndpoint macro

AgapovOne:
I'm attempting to fix behavior that broke after #355.

cc @mbrandonw

---

Still in draft because it would have a possible bug that were removed from documentation in release 1.9.2:

<details>
<summary>Old bug</summary>

Screenshot 2025-04-23 at 11 28 40

</details>

I'm open to ideas on how we can fix this. 

If we can't, probably asking everyone to provide own init is better than have this bug still here.

The library works fine with all old cases, and I just found one specific case that would break it, but it can be avoided with custom init.

### DIFF
--- a/Sources/DependenciesMacros/Macros.swift
+++ b/Sources/DependenciesMacros/Macros.swift
@@ -204,7 +204,7 @@ public macro DependencyClient() =
 /// ```
 @attached(accessor, names: named(init), named(get), named(set))
 @attached(peer, names: arbitrary)
-public macro DependencyEndpoint(method: String = "") =
+public macro DependencyEndpoint(method: String = "", initAccessor: Bool = false) =
   #externalMacro(
     module: "DependenciesMacrosPlugin", type: "DependencyEndpointMacro"
   )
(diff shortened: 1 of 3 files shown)


mbrandonw:
Hi @AgapovOne, I am starting to feel that even allowing the `initAccessor` option is maybe too much of a footgun. I just don't think people should be using `@DependencyEndpoint` directly. It really doesn't do that much for you. Instead of this:

```swift
@DependencyEndpoint
public var set: (ValueType?) -> Void
```

…you do this:

```swift
public var set: (ValueType?) -> Void = unimplemented("Unimplemented: \(Self.self).set")
```

So, while our change was technically a breaking change, there is a clear migration path forward, and I think we want to overall discourage people from using `@DependencyEndpoint`. We should probably deprecate it and replace it with `@_DependencyEndpoint`.

What do you think?

AgapovOne:
I think we shouldn't bring my change into repository because it brings again this issue with missing `reportIssue` call.

But I want to explain where I'm using `@DependencyEndpoint` now.

```
@DependencyEndpoint
public var set(_ value: ValuType?) -> Void
```

Brings me named arguments. That's the only thing I'm using it for. 

```
set(value: nil)
```

It's probably better to write it myself now. Or look for other macro.

---

> deprecate it and replace it with @_DependencyEndpoint.

The point of this would be to discourage it's use without `@DependencyClient`?

PR #363, opened 2025-04-24
Fix incorrect use of `#line` where `#column` was intended

omochi:
It should be `#column`, right?

I found two instances.
I searched using the method shown in the screenshot below, so I don’t think there are any others.

スクリーンショット 2025-04-24 10 28 06


mbrandonw (approved):
Oops, good catch. Thanks!

PR #366, opened 2025-04-30
TestTrait throwing fix for pre-6.1

stephencelis:
We only fixed the `throws` availability on 6.1+.


mbrandonw:
Closing this because trailing closures don't work for traits in Swift <6.1.

PR #376, opened 2025-06-24
Try importing `Musl` if `Glibc` isn't available

slice:
This improves platform support by letting the library be compiled by e.g. the Swift Static Linux SDK. The import pattern follows [an example][example] presented in the Swift docs.

[example]: https://www.swift.org/documentation/articles/static-linux-getting-started.html#:~:text=If%20you%20are%20using%20such


stephencelis (approved):
Thanks!

PR #377, opened 2025-07-16
Prefer `#if compiler` over `#if swift`

jflan-dd:
I ran into a compiler error when compiling this package using the Swift 6.1 compiler in Swift 5 language mode.

The specific error I hit was caused by the mismatch between `#if swift(>=6.1)` check on [`_DependencyTrait`](https://github.com/pointfreeco/swift-dependencies/blob/7dc03704f892cfdaa22ac3213fed5ce6bf8a831f/Sources/DependenciesTestSupport/TestTrait.swift#L6) and the `if compiler(<6.1)` check around [`testValuesByTestID`](https://github.com/pointfreeco/swift-dependencies/blob/7dc03704f892cfdaa22ac3213fed5ce6bf8a831f/Sources/Dependencies/Traits/TestTrait.swift#L1).

That resulted in not being able to compile the following code with the Swift 6.1 compiler in Swift 5 language mode because `testValuesByTestID` wasn't defined.
```swift
    extension _DependenciesTrait: SuiteTrait, TestTrait {
      public var isRecursive: Bool { true }

      public func prepare(for test: Test) async throws {
        try testValuesByTestID.withValue {
          try self.updateValues(&$0[test.id, default: DependencyValues(context: .test)])
        }
      }
    }
```

We're using Bazel to depend on this package, which is how we're hitting this issue, but you can replicate it with SPM by running `swift test -Xswiftc -swift-version -Xswiftc 5` in this repo.

I updated every instance of `#if swift` to `#if compiler` in this PR and had passing tests with both `swift test` and `swift test -Xswiftc -swift-version -Xswiftc 5`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         config: ['debug', 'release']
-        xcode: ['16.2']
+        xcode: ['16.4']
     steps:
       - uses: actions/checkout@v4
       - name: Select Xcode ${{ matrix.xcode }}


stephencelis:
@jflan-dd Thanks for looking into this!

Did you change every `#if swift` to `#if compiler`, or are these all compiler features being bracketed, and not language mode features?

If the former, my worry is that this may introduce unintended changes where `#if swift` _was_ the appropriate thing to do. Do you think you could update this PR to simply make the required change(s) for your app to build successfully?

If you do believe all of these changes are appropriate, could you comment on each `#if` changed with your reasoning? We just want to understand things better.

jflan-dd:
@stephencelis I commented on the changes in `Sources`. I reverted one change that depends on a Swift 6 language mode feature, but I believe the remaining two are valid changes.

I built and ran tests after deleting the `Package@swift-6.0.swift` file to force using Swift 5 language mode and didn't have any new warnings after reverting the check related to KeyPath sendability.

The instances in `Tests` looked like they were all about TestScoping.

Sources/Dependencies/WithDependencies.swift
```diff
@@ -314,7 +314,7 @@ public func withDependencies<Model: AnyObject, R>(
   )
 }
 
-#if swift(>=6)
+#if compiler(>=6)
```
jflan-dd, line 317: also #isolation

jflan-dd:
@stephencelis Do the updated comments make sense?

stephencelis:
@jflan-dd Yup! Just lost track of this, sorry :smile:

For the key path sendability, want to add the feature flag to the Swift 5 package.swift and see if that helps things?

stephencelis:
@jflan-dd Thanks for this! All good to go!

PR #383, opened 2025-09-03
Add @DependencyMock property wrapper for safe mock casting in tests

maatheusgois-dd:
- Provides type-safe access to mock dependencies in test contexts
- Includes helpful error messages when types don't match
- Only available in test contexts (Testing/XCTest)
- Follows library conventions with Sendable compliance

### DIFF
--- a/Sources/DependenciesTestSupport/DependencyMock.swift
+++ b/Sources/DependenciesTestSupport/DependencyMock.swift
@@ -1,0 +1,52 @@
+#if canImport(Testing) || canImport(XCTest)
+  import Foundation
+  import Dependencies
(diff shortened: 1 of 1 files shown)


stephencelis:
@maatheusgois-dd Whoops, I was re-reviewing PRs that have gone by the wayside and looks like the same functionality was implemented in #413 with a slightly different shape. Will credit you for the idea in the release notes! Sorry for not letting you steer things in this PR 😅

mbrandonw:
Sorry we lost track of this @maatheusgois-dd! But glad it eventually made it into the library :)

PR #389, opened 2025-09-17
Support SwiftSyntax 602

mbrandonw:


Sources/Dependencies/Internal/Deprecations.swift
```diff
@@ -105,13 +105,14 @@ extension AsyncThrowingStream where Failure == Error {
 
 // MARK: -
 
+@available(*, deprecated)
```
mbrandonw, line 108: A few unrelated changes to get rid of some deprecation warnings in the project/tests.

stephencelis (approved):

PR #414, opened 2026-02-01
Fix UUID example strings in documentation

bradhowes:


Sources/Dependencies/DependencyValues/UUID.swift
```diff
@@ -118,14 +118,14 @@ extension UUID {
...
-  /// UUID(16) == UUID(uuidString: "00000000-0000-0000-0000000000F0")
+  /// UUID(16) == UUID(uuidString: "00000000-0000-0000-0000-000000000010")
   /// ```
   ///
   /// If a negative number is passed to this function then it is inverted and the negative sign
   /// is encoded into the 16th bit of the UUID:
   ///
   /// ```swift
-  /// UUID(-16) == UUID(uuidString: "00000000-0001-0000-0000000000F0")
+  /// UUID(-16) == UUID(uuidString: "00000000-0000-0001-0000-000000000010")
```
pyrtsa: This negative number should actually end with `F0`:
```suggestion
  /// UUID(-16) == UUID(uuidString: "00000000-0000-0001-0000-0000000000F0")
```

bradhowes: let uat = UUID(-16)
print(uat.uuidString). ==> "00000000-0000-0001-0000-000000000010"

PR #424, opened 2026-02-19
Bump minimal support to Swift 6.1

stephencelis:
We technically support 5.9 now, but folks probably shouldn't be running that version of Swift still, and if they must they can lock their version of Dependencies. This allows us to prune a bunch of old branches of code.


davdroman:
@stephencelis this is also superseded by #462 btw :)

stephencelis:
@davdroman 🫡

PR #438, opened 2026-05-11
Skip URLSession on Android to drop libFoundationNetworking from DT_NEEDED

mansbernhardt:
Wraps the outer `#if Foundation` in `URLSession.swift` with `&& !os(Android)`, skipping the file entirely when cross-compiling for the Swift Android SDK. Why this file specifically: the `UnimplementedURLProtocol: URLProtocol` subclass near the bottom registers `URLProtocol` class metadata at module init. On Android, `URLProtocol` lives in `libFoundationNetworking.so`, so the metadata reference forces the linker to add that lib to the consumer binary's `DT_NEEDED` list — ~16 MB of dead weight for projects that never reach for `\.urlSession` and would otherwise drop the lib entirely. No source change outside the `#if` guard. No behavior change on Apple, Linux, WASM, or any other target — `canImport(Android)` is false everywhere else. The `\.urlSession` dependency value remains available on every other platform that has the Foundation trait enabled.
Consumers that *do* need `\.urlSession` on Android can re-introduce it downstream behind their own gate, or pull `URLSession` from `FoundationNetworking` directly.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -29,12 +29,6 @@ let package = Package(
     "Clocks",
     "CombineSchedulers",
     "Foundation",
-    // Gates the `\.urlSession` dependency value (defined in
-    // `DependencyValues/URLSession.swift`). When disabled, the `UnimplementedURLProtocol:
-    // URLProtocol` subclass inside that file is excluded from the build, so consumers
-    // cross-compiling against the Swift Android SDK no longer pull
-    // `libFoundationNetworking.so` (~16 MB) into their bridge's `DT_NEEDED`. Default on
-    // for backward compatibility — explicit opt-out via `traits:` on `.package(...)`.
     "FoundationNetworking",
     .default(enabledTraits: [
       "Clocks", "CombineSchedulers", "Foundation", "FoundationNetworking"
--- a/Package.swift
+++ b/Package.swift
@@ -29,8 +29,15 @@ let package = Package(
     "Clocks",
     "CombineSchedulers",
     "Foundation",
+    // Gates the `\.urlSession` dependency value (defined in
+    // `DependencyValues/URLSession.swift`). When disabled, the `UnimplementedURLProtocol:
+    // URLProtocol` subclass inside that file is excluded from the build, so consumers
+    // cross-compiling against the Swift Android SDK no longer pull
+    // `libFoundationNetworking.so` (~16 MB) into their bridge's `DT_NEEDED`. Default on
+    // for backward compatibility — explicit opt-out via `traits:` on `.package(...)`.
+    "FoundationNetworking",
     .default(enabledTraits: [
-      "Clocks", "CombineSchedulers", "Foundation"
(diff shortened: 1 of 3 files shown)


mbrandonw:
Hi @mansbernhardt, thanks for investigating this. I think a better change would be to introduce a dedicated `FoundationNetworking` trait for this instead of including `URLSession` with just the `Foundation` trait. In order to make it backwards compatible you could include `URLSession` when `FoundationNetworking` is enabled _or_ when on Apple platforms. Can you give that a shot?

mansbernhardt:
Sounds good, I have pushed an update adding a FoundationNetworking trait.

Sources/Dependencies/DependencyValues/URLSession.swift
```diff
@@ -1,4 +1,13 @@
-#if Foundation
+// Gated on the `FoundationNetworking` package trait (default on) OR Apple platforms.
+// The `UnimplementedURLProtocol: URLProtocol` subclass below registers `URLProtocol`
+// class metadata at module init time. On split-Foundation platforms (Linux, Android,
+// swift-corelibs-foundation) that pulls `libFoundationNetworking.so` (~16 MB) into a
+// consumer's binary `DT_NEEDED` list. Apple platforms always include the value (the
+// trait is a no-op on Darwin since `URLSession` lives in `Foundation` proper there).
+// Split-Foundation consumers that don't need `\.urlSession` can opt out by declaring
+// `traits: ["Clocks", "CombineSchedulers", "Foundation"]` on their `.package(...)`
+// (omitting `FoundationNetworking`).
```
mbrandonw:
```suggestion
```

Package.swift
```diff
@@ -29,8 +29,15 @@ let package = Package(
     "Clocks",
     "CombineSchedulers",
     "Foundation",
+    // Gates the `\.urlSession` dependency value (defined in
+    // `DependencyValues/URLSession.swift`). When disabled, the `UnimplementedURLProtocol:
+    // URLProtocol` subclass inside that file is excluded from the build, so consumers
+    // cross-compiling against the Swift Android SDK no longer pull
+    // `libFoundationNetworking.so` (~16 MB) into their bridge's `DT_NEEDED`. Default on
+    // for backward compatibility — explicit opt-out via `traits:` on `.package(...)`.
```
mbrandonw:
```suggestion
```

mbrandonw (approved):
Thanks!

Sources/Dependencies/DependencyValues/URLSession.swift
```diff
@@ -1,4 +1,4 @@
-#if Foundation
+#if FoundationNetworking || canImport(Darwin)
```
mbrandonw:
```suggestion
#if FoundationNetworking
```

mbrandonw (approved):
Thanks!

PR #439, opened 2026-05-12
Fix Linux CI.

mbrandonw:


Sources/Dependencies/Internal/TypeName.swift
```diff
@@ -1,3 +1,5 @@
+import Foundation
```
mbrandonw, line 1: We just either use the `Foundation` trait for this or stop using Foundation fort this helper.

Package.swift
```diff
@@ -63,6 +63,9 @@ let package = Package(
         .product(name: "ConcurrencyExtras", package: "swift-concurrency-extras"),
         .product(name: "IssueReporting", package: "xctest-dynamic-overlay"),
         .product(name: "XCTestDynamicOverlay", package: "xctest-dynamic-overlay"),
+      ],
+      swiftSettings: [
+        .enableUpcomingFeature("MemberImportVisibility")
```
mbrandonw, line 69: We should start enabling this across all of our packages.

stephencelis (approved):

Younglilly4 (approved):
Please go on

Younglilly4 (approved):
Please go on

PR #442, opened 2026-06-01
Fix stale documentation links

davdroman:
## Summary
- Replace stale pointfreeco.github.io DocC links with Swift Package Index documentation links.
- Update the Issue Reporting unimplemented link to the current SPI documentation path.

## Verification
- Checked that no files still reference pointfreeco.github.io.
- Ran git diff --check.


stephencelis (approved):
Thanks!

davidcairns:
@stephencelis @davdroman May be worth noting that the new URL is currently giving me a 404 (but the old one works fine lol)

davdroman:
Same here. SPI docs seems to be down for Dependencies for some reason. I thought I got the link wrong but in fact if you [go here](https://swiftpackageindex.com/pointfreeco/swift-dependencies) and click "Documentation" it 404s.

PR #451, opened 2026-06-17
Adds the `FoundationNetworking` trait to Package@swift-6.0.swift

rhysm94:
Resolves [the discussion from Slack](https://pointfreecommunity.slack.com/archives/C04L2D0MNJH/p1781733403850059) where I was encountering a failed compile.
This trait was missing from Package@swift-6.0.swift, so older Xcode versions were missing the `\.urlSession` dependency key.


mbrandonw:
Thanks!

PR #456, opened 2026-07-09
Add Swift 6.1 package manifest

davdroman:
As discussed in https://github.com/pointfreeco/swift-sharing/pull/215#issuecomment-4879104261, this adds `Package@swift-6.1.swift` so Swift 6.1 and Swift 6.2 toolchains can resolve the package while the primary manifest uses Swift tools 6.3.

The manifest mirrors the current package setup, including package traits, and only enables upcoming features available to Swift 6.1.

## Verification

Selected Xcode 26.3 and ran:

- `make test-swift`
- `swift package dump-package | jq '.toolsVersion'` confirmed the selected manifest reports tools version `6.1.0`

### DIFF
--- a/Package@swift-6.1.swift
+++ b/Package@swift-6.1.swift
@@ -1,0 +1,149 @@
+// swift-tools-version: 6.1
+
+import CompilerPluginSupport
+import PackageDescription
+
+let package = Package(
+  name: "swift-dependencies",
+  platforms: [
+    .iOS(.v13),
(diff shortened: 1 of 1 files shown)


stephencelis:
@davdroman Sorry, missed this! Lot of work inflight across many libraries, but I believe #462 supersedes this, and there's no need to maintain both 6.3 and 6.1 package files.

